PR #6, opened 2025-06-19
Add clsx + fix some minor UI

HocusLocusTee:
## What?

Add [`clsx`](https://www.npmjs.com/package/clsx) so we can move to utility based Tailwind for UI styling (conditionals etc).

Misc UI improvements:
- Remove some css and replace with Tailwind, includes restyling the buttons
   - This has also removed a weird janky border around the sub header
   - Buttons touched are: Start Wallet Services, Wallet Info, Close Wallet
- Add connected network name and wallet name to main page
- Move close wallet and start wallet services to the right, replacing them with network and wallet info
- Add cursor-pointer to 2 of the main buttons (which oddly are getting blocked while we load messages, to be looked at)

_Note: the selection of the above was solely based on stuff that annoyed me on my initial use._

Images:

S1
S2

### DIFF
--- a/src/OneLiner.tsx
+++ b/src/OneLiner.tsx
@@ -275,7 +275,11 @@ export const OneLiner: FC = () => {
  <div className="px-8 py-4 bg-[var(--primary-bg)]">
   <div className="flex items-center gap-4">
     {isWalletReady ? (
-      <div className="flex flex-row justify-between items-start gap-4 w-full">
+      <div className="flex flex-col sm:flex-row justify-between items-start gap-4 w-full"> 
+        <div className="flex flex-col items-start text-xs gap-1 whitespace-nowrap">
+          <div><strong>Network:</strong> {walletStore.selectedNetwork}</div>
+          <div><strong>Wallet Name:</strong> {unlockedWalletName}</div>
+        </div>
         <div className="flex items-center gap-2 text-sm">
           <button
             className={clsx(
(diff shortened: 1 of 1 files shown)


src/OneLiner.tsx
```diff
@@ -264,44 +267,53 @@ export const OneLiner: FC = () => {
...
-            </>
-          ) : (
-            <WalletGuard
-              onSuccess={onWalletUnlocked}
-              selectedNetwork={selectedNetwork}
-              onNetworkChange={setSelectedNetwork}
-              isConnected={isConnected}
-            />
-          )}
+ <div className="px-8 py-4 bg-[var(--primary-bg)]">
```
ABitShift, line 275: maybe we should start including color names into `tailwind.config.js` (could also be swapped to `tailwind.config.ts`).

this would change the color use from `-[var(--primary-bg)]` to `-primary-bg`

src/OneLiner.tsx
```diff
@@ -264,44 +267,53 @@ export const OneLiner: FC = () => {
...
+    {isWalletReady ? (
+      <div className="flex flex-col sm:flex-row justify-between items-start gap-4 w-full"> 
+        <div className="flex flex-col items-start text-xs gap-1 whitespace-nowrap">
+          <div><strong>Network:</strong> {walletStore.selectedNetwork}</div>
+          <div><strong>Wallet Name:</strong> {unlockedWalletName}</div>
+        </div>
+        <div className="flex items-center gap-2 text-sm">
+          <button
+            className={clsx(
+              "bg-[var(--accent-blue)] hover:bg-[var(--accent-blue)]/90 text-white text-sm font-bold py-2 px-4 rounded cursor-pointer",
```
ABitShift, line 286: same about `accent-blue`

HocusLocusTee: Good ideas. 
I kept these `as is` because its already used on a few other CSS things and we could just standardize it once its all cut over to Tailwind.

PR #11, opened 2025-06-20
Add the ability to override minimum password length

HocusLocusTee:
## What?

This PR takes the current crude password length logic and sets it based on an environment variable.
Currently set to `false` in the staging and production .env files.

You can turn the password length off by adding:
`VITE_DISABLE_PASSWORD_REQUIREMENTS=true` to your `.env.local` file.

Additionally, I created a `password.ts` file in `/config/`. This is anticipating some potentially more complex password requirements over time.


src/containers/WalletGuard.tsx
```diff
@@ -122,10 +123,12 @@ export const WalletGuard = ({
...
-    if (passwordRef.current!.value.length < PASSWORD_MIN_LENGTH) {
-      setError("Password must be at least 4 characters");
-      return;
-    }
+    const pw = passwordRef.current!.value;
+if (!disablePasswordRequirements && pw.length < PASSWORD_MIN_LENGTH) {
+  setError(`Password must be at least ${PASSWORD_MIN_LENGTH} characters`);
+  return;
+}
+
```
IzioDev, line 131: should we start introducing a workspace specific vscode configuration file to globally fix indent?

HocusLocusTee: Yep, Ill get onto it soon.

For some reason I thought I saw a commit from @AuzGhosty starting this.

PR #29, opened 2025-06-21
feat: add logging based on env

HocusLocusTee:
## What?
Adds console.log function overrides.

By default all of our console logs, warns and errors will now be turned off in production and on in staging. This is done through env variables.

For devs, youll need to add `VITE_ALLOW_VERBOSE_LOGGING=false` to you `.env.local` file.


IzioDev:
Awesome!
I could see a case where an end-user experience a bug and want to share some details (such as where an error occurred).

In this context, do you think we could add a layer of logic on `LOG_LEVEL`, selectively accepting or rejecting a write on console?

For example, given the following log levels:
* INFO
* WARN
* ERROR
I can set my `LOG_LEVEL` to `WARN` and only see console writes if the `LEVEL >= WARN` (`WARN` and `ERROR` in this case)

Issue #19, opened 2025-06-21
UI component modularity and mobile responsiveness considerations


IzioDev:
The application codebase initially was a native Javascript application. We chose to initiate a quick transition to React, knowing we would need to spend more effort to completely transition to a React paradigm.

This issue will be a long ongoing effort into:
1. modularization of the existing component (especially OneLiner.tsx)
2. starting to drop CSS in favor of TailwindCSS (inline)
3. considering mobile responsiveness
    * some UI element aren't mobile friendly (such as modal), and might need deeper efforts and discussions in order to achieve this point

IzioDev:
One point raised by @Ernie is that errors and alerts displays aren't uniforms, and we should make them iso

HocusLocusTee:
I'm starting to work on this.

PR #33, opened 2025-06-21
Styling - OneLiner and some mobile stuff

HocusLocusTee:
## What?
Work toward [Issue 19](https://github.com/K-Kluster/Kasia/issues/19) this PR removes all css from OneLiner.

Functionality should be identical except for the following changes:
Main chat window now takes up 70vh instead of a fixed height - its minimum is the same

Additionally I did some small mobile enhancements:
- Contact Card - Fixed nickname buttons, allowed address / nick to be handled properly, previously they spilled
- Fee Estimates go into the hamburger on mobile. NOTE: Maybe we should eventually have an alter that persists next to the hamburger IF the fee is higher than base? But that's not in this PR
- Centered the network / wallet info and see address button on mobile

Also: Hamburger menu now closes when you click outside of it

Mobile: will probably get a big redesign at some point but my current goal is to get the current design working ok on mobile

**See discord for quick video of changes**

### DIFF
--- a/src/OneLiner.tsx
+++ b/src/OneLiner.tsx
@@ -230,8 +230,8 @@ export const OneLiner: FC = () => {
       <div className="px-8 py-4 bg-[var(--primary-bg)]">
         <div className="flex items-center gap-4">
           {isWalletReady ? (
-            <div className="flex flex-col sm:flex-row justify-between items-start gap-4 w-full">
-              <div className="flex flex-col items-start text-xs gap-1 whitespace-nowrap">
+            <div className="flex flex-col sm:flex-row sm:justify-between sm:items-start items-center gap-4 w-full text-xs">
+              <div className="flex flex-col items-start sm:items-start items-center gap-1 whitespace-nowrap">
                 <div>
                   <strong>Network:</strong> {walletStore.selectedNetwork}
                 </div>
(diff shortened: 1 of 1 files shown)


src/components/WalletInfo.tsx
```diff
@@ -116,8 +116,10 @@ export const WalletInfo: FC<WalletInfoProps> = ({
   if (!isWalletReady || !open) return null;
 
   return (
-    <div className="modal-overlay" onClick={onClose}>
-      <div className="modal-content" onClick={(e) => e.stopPropagation()}>
+    <div className="fixed inset-0 bg-black/50 flex items-center justify-center z-[1000]"
+      onClick={onClose}>
+      <div className="bg-[var(--secondary-bg)] p-5 rounded-xl relative max-w-[500px] w-[90%] max-h-[90vh] overflow-y-auto border border-[var(--border-color)] animate-[modalFadeIn_0.3s_ease-out]"
+       onClick={(e) => e.stopPropagation()}>
```
IzioDev, line 122: is it needed?

HocusLocusTee: Nope we need it, it keeps the modal open if you click INSIDE of it.

PR #39, opened 2025-06-21
style: replace css with tailwind + svg with herocn

HocusLocusTee:
## What?
This PR removes css and replaces it with Tailwind in our `WalletAddressSection` component.

I've removed some of the key frame stuff and replaced it with opacity transitions.
Also, I've replaced some SVGs with Heroicons.

Other than that, it essentially looks the same.


IzioDev:
Tested locally, LGTM!
Thanks for this contribution, again!

PR #46, opened 2025-06-23
feat: ui & ux improvements

IzioDev:
I've made a few changes:
1. auto-focus on password when selected a wallet to unlock
2. remove the kaspa address above the message you enter, this isn't necessary to have that there anymore
3. changed the box title `Messages` with the address you're talking to (useful when the contact name isn't the address):
  * possibility to click (...) in order to expend the address
  * possibility to click clipboard icon to copy the address
  => note: i don't like the icon, not sure if it's the placement or else but it sucks as is i believe
4. auto focus the address input when creating a new conversation
5. keep the active conversation open when creating a new conversation. Previously, when the modal opened, Kasia would have closed the opened conversation you had

Here is a few things i also wanted to address but i didn't have time:
* change the fee estimation display, not sure how yet but at least makes it less frustrating when entering long text
* shift+enter => insert new line instead of sending the message


HocusLocusTee:
A few (non-code) things while I'm on my branch that is base on this:

Need to adjust the z of the copied tooltip.
image

When trying manual connection, we lock out the save button to *try* and connect...  if I choose to go back - do you think this should cancel that attempt?

HocusLocusTee:
>change the fee estimation display, not sure how yet but at least makes it less frustrating when entering long text

Moving above the text box will help out of the box. But once we start getting "multi-line" style textareas - the user might miss it.

PR #57, opened 2025-06-23
KNS Implementation

atwci:
+ Create new chat by using KNS domain to resolve address and sets nickname as the KNS domain
+ If KNS domain changes owners it'll let the user know

### DIFF
--- a/src/components/NewChatForm.tsx
+++ b/src/components/NewChatForm.tsx
@@ -1,95 +1,154 @@
-import React, { useState, useCallback, useEffect } from 'react';
-import { useMessagingStore } from '../store/messaging.store';
-import { useWalletStore } from '../store/wallet.store';
-import { kaspaToSompi, sompiToKaspaString } from 'kaspa-wasm';
-import styles from './NewChatForm.module.css';
+import React, { useState, useCallback, useEffect, useRef } from "react";
+import { useMessagingStore } from "../store/messaging.store";
+import { useWalletStore } from "../store/wallet.store";
+import { kaspaToSompi, sompiToKaspaString } from "kaspa-wasm";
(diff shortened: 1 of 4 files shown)


src/components/NewChatForm.tsx
```diff
@@ -1,95 +1,154 @@
...
+          setRecipientWarning(
+            "Could not verify recipient balance. They may not be able to respond if they have no KAS."
+          );
+          return;
+        }
+        const balanceData = await response.json();
+        const balance = BigInt(balanceData.balance || 0);
+        if (balance === BigInt(0)) {
+          setRecipientWarning(
+            "⚠️ Warning: Recipient has zero KAS balance and will not be able to respond to your handshake. Consider sending a higher amount."
```
HocusLocusTee, line 126: It's not covered in the docs yet (sorry!) but we are trying to move forward with HeroIcons only.

src/components/NewChatForm.tsx
```diff
@@ -1,95 +1,154 @@
...
+      setIsCheckingRecipient(true);
+      setRecipientWarning(null);
+      try {
+        const networkId = walletStore.accountService?.networkId || "mainnet";
+        const baseUrl =
+          networkId === "mainnet"
+            ? "https://api.kaspa.org"
+            : "https://api-tn10.kaspa.org";
+        const encodedAddress = encodeURIComponent(address);
+        const response = await fetch(
```
HocusLocusTee, line 113: Going forward, I think we should be extracting the core fetch req out into a helper lib.
Wdyt @IzioDev ?

IzioDev:
Please consider this branch as outdated and move to #63

PR #66, opened 2025-06-24
feat/mobile improvements, payment field

HocusLocusTee:
## What?

This introduces the ability to show and hide the chat panel, adding an avatar when collapsed. The avatar contains the first letter of the nickname or the first letter of the KAS address.
Additionally, we have a focused panel state on mobile! You can either view the chat or messages, not both (like nearly every mobile chat app). I've also moved delete to the actions list.

A more neutral change is in the send message bar: I've placed both the payment and file attach into a popover. This works nicely, but I struggled to get the payment popover (it's also a popover, ha) to sit nicely. So for now it's been hacked into place. Also, I've adjusted the payment popover to sit a bit nicer on mobile.

There might be some slight, annoying state bugs if you're on a computer dragging between the two (mobile and desktop) states, but they should all be minor.

Probably needs some fine-tuning, but it's a lot of the way there.

Oh, and I removed some CSS and gave the initial loading message client state a small pulse.


IzioDev:
It keeps getting and better, looks awesome, LGTM!

PR #69, opened 2025-06-25
@username system using KNS

atwci:
+ Fixed previous missing code
+ UI uses the already existing colors now

### DIFF
--- a/src/components/NewChatForm.tsx
+++ b/src/components/NewChatForm.tsx
@@ -351,15 +351,11 @@ export const NewChatForm: React.FC<NewChatFormProps> = ({
       (!isKaspaAddress && recipientAddress.length > 0);
 
     if (isKnsDomain && resolvedAddress) {
-      // Convert to @username format for display
-      let displayName = recipientAddress;
+      let username = recipientAddress;
       if (recipientAddress.endsWith(".kas")) {
-        displayName = "@" + recipientAddress.slice(0, -4); // Remove .kas and add @
-      } else {
-        displayName = "@" + recipientAddress; // Add @ to username
+        username = recipientAddress.slice(0, -4); // Remove .kas
       }
-
-      recipientDisplay = displayName;
+      recipientDisplay = `@${username} (${username}.kas)`;
     } else {
       recipientDisplay = knsRecipientAddress;
     }
@@ -445,7 +441,8 @@ export const NewChatForm: React.FC<NewChatFormProps> = ({
             <div className={styles["input-container"]}>
               {!recipientAddress.startsWith("kaspa:") &&
                 !recipientAddress.startsWith("kaspatest:") &&
-                recipientAddress.length > 0 && (
+                recipientAddress.length > 0 &&
+                !recipientAddress.endsWith(".kas") && (
                   <span className={styles["at-prefix"]}>@</span>
                 )}
               <input
@@ -453,7 +450,8 @@ export const NewChatForm: React.FC<NewChatFormProps> = ({
                 className={`${styles.input} ${
                   !recipientAddress.startsWith("kaspa:") &&
                   !recipientAddress.startsWith("kaspatest:") &&
-                  recipientAddress.length > 0
+                  recipientAddress.length > 0 &&
+                  !recipientAddress.endsWith(".kas")
                     ? styles["username-input"]
                     : ""
                 }`}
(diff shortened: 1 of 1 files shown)


IzioDev:
Currently updating the base branch `upstream/atwci-KNS` from staging.
Fixing conflicts, and then we can progress with this :)

Thanks for contributing again!

src/components/NewChatForm.tsx
```diff
@@ -2,28 +2,51 @@ import { useMessagingStore } from "../store/messaging.store";
...
 import React, { useState, useCallback, useEffect, useRef } from "react";
 import { kaspaToSompi, sompiToKaspaString } from "kaspa-wasm";
+import { KaspaAddress } from "./KaspaAddress";
 import styles from "../components/NewChatForm.module.css";
 
 interface NewChatFormProps {
   onClose: () => void;
+  prefilledRecipient?: string;
+  prefilledResolvedAddress?: string;
+  prefilledDomainId?: string;
```
IzioDev, line 12: these props are defined but never used

src/components/NewChatForm.tsx
```diff
@@ -2,28 +2,51 @@ import { useMessagingStore } from "../store/messaging.store";
...
 
-export const NewChatForm: React.FC<NewChatFormProps> = ({ onClose }) => {
-  const [recipientAddress, setRecipientAddress] = useState("");
+export const NewChatForm: React.FC<NewChatFormProps> = ({
+  onClose,
+  prefilledRecipient = "",
+  prefilledResolvedAddress = "",
+  prefilledDomainId = "",
+}) => {
+  const [recipientAddress, setRecipientAddress] = useState(prefilledRecipient);
```
IzioDev, line 21: recipient address should be renamed to `recipientValue` since it can contains either an address or a KNS domain

src/components/NewChatForm.tsx
```diff
@@ -2,28 +2,51 @@ import { useMessagingStore } from "../store/messaging.store";
...
-  const [knsDomainId, setKnsDomainId] = useState<string | null>(null);
+  const [knsDomainId, setKnsDomainId] = useState<string | null>(
+    prefilledDomainId || null
+  );
   const knsDomainRef = useRef<string>("");
 
+  // Helper to determine if input is a Kaspa address
+  const isKaspaAddress =
+    recipientAddress.startsWith("kaspa:") ||
+    recipientAddress.startsWith("kaspatest:");
```
IzioDev, line 45: this can be memoized:
```ts
const isRecipientValueAnAddressFormat = useMemo(() => {
  return (
    // expression here
  )
}, [recipientAddress])

File: src/components/NewChatForm.tsx [hunk 4]
```diff
@@ -2,28 +2,51 @@ import { useMessagingStore } from "../store/messaging.store";
...
+
+  // Helper to determine if input could be a username
+  const couldBeUsername = !isKaspaAddress && recipientAddress.length > 0;
```
IzioDev, line 48: We now use two sources of truth for detecting the format: `isKaspaAddress` (proposed to be renamed: `isRecipientValueAnAddressFormat`) and  `couldBeUsername`

Proposal to unify these two with:
```ts
const detectedRecipientFormat = useMemo<'address' | 'kns' | 'undetermined'>(() => {
  // logic here
}, [recipientAddress])
```

src/components/NewChatForm.tsx
```diff
@@ -61,13 +84,35 @@ export const NewChatForm: React.FC<NewChatFormProps> = ({ onClose }) => {
...
   // KNS domain resolution effect
   useEffect(() => {
-    if (recipientAddress.endsWith(".kas")) {
+    // Only check for KNS domains if input could be a username
+    if (!couldBeUsername) {
+      setResolvedAddress(null);
+      setKnsDomainId(null);
+      setKnsError(null);
+      return;
+    }
```
IzioDev, line 93: can now be optimized by relying of `format` effect

IzioDev (commented):
I did not finish the review, I'm making changes myself if that's ok for you.

The reason behind it is that we would like to benefit from this feature the sooner so we can integrate it on `staging` and begin tests on it.

I still keep the pending comment as a reference for you so you get the idea. You'll get pinged once changes has been made. Feel free to notify me, or whoever, if you disagree with implemented changes, open for discussion!

PR #77, opened 2025-06-26
docs: brief design primer

HocusLocusTee:
## What?
Some brief but not exhaustive set of notes re ui design


IzioDev:
This is awesome!

Issue #79, opened 2025-06-26
Configurable fee thresholds & hard-confirm for high fees


HocusLocusTee:
Current fee color buckets use arbitrary limits (implemented [here](https://github.com/K-Kluster/Kasia/pull/80)). 

Follow up work:

- Compute sensible default thresholds
- Expose thresholds in Settings for user override
- Add extra “Confirm send” step when estimated fee exceeds the high-fee threshold (0.05 KAS?)

HocusLocusTee:
Done.

PR #80, opened 2025-06-26
Feat/fee estimate

HocusLocusTee:
## What?

Moves the fee estimate from below the chat line (and shifting layout) to above.
Adds dynamic styling based on fee amount.
Deletes some css

Issue raised [here ](https://github.com/K-Kluster/Kasia/issues/79)) for further work.


src/containers/SendMessageForm.tsx
```diff
@@ -19,9 +19,23 @@ import {
...
+  { limit: 0.00002000, classes: "text-green-400 border-green-400" },
+  { limit: 0.00005000, classes: "text-blue-400  border-blue-400" },
+  { limit: 0.0005, classes: "text-yellow-400 border-yellow-400" },
+  { limit: 0.001, classes: "text-orange-400 border-orange-400" },
+  { limit: Infinity, classes: "text-red-400 border-red-400" },
+];
+
+function getFeeClasses(fee: number) {
+  return FEE_LEVELS.find(({ limit }) => fee <= limit)!.classes;
+}
```
IzioDev, line 37: Topic performance:

`feeClasses` should be memoized from the (only) `feeEstimate` dependency:
```ts
const estimatedFeesDisplayClasses = useMemo(() => {
    return FEE_LEVELS.find(({ limit }) => fee <= limit)!.classes;
}, [feeEstimate])
```

IzioDev (changes requested):
Besides this, LGTM!

src/containers/SendMessageForm.tsx
HocusLocusTee: Seems negligible for sure a small array.
But good standard never the less.

PR #83, opened 2025-06-26
Fix for UTXO compounding issue

pratamatama:
The safeguard attempt to multiple transaction turns out to be the cause of the unnecessarily complex compounding logic. Removing that part makes it simpler and allows us to naturally perform the compound transaction.

This PR also removes the no longer relevant elements related to the batching process.


IzioDev (changes requested):
LGTM, just need a removal of the comments and we're ready to ship this!

Thank you for your contribution! ❤️

PR #89, opened 2025-06-27
Feat/routing refactor styling

HocusLocusTee:
## What?
amongst a bunch of other stuff, this adds some sensible routing. It is a big refactor of the current code and layout.

- `/`  
- `/wallet/create`  
- `/wallet/import`  
- `/wallet/unlock/:wallet`  
- `/wallet/migrate/:wallet`  
- `/:walletId`  
- `/settings-network`  

addresses [this issue.](https://github.com/K-Kluster/Kasia/issues/71)

_note: again, some of these rewritten still have high potential to be refactored. eg walletguard. But this PR is big enough so I just ''made things work" for now._

### changes
- strip out a bunch of stuff from oneliner, menus, logic, walletguard etc  
- modify walletguard to route based on step, also some extra bits to be sure passwords are clear  
- connect at top app level  
- modify app.tsx and add routes  
- refactor all menus and bring into a rootlayout, store menu open state (maybe bring modals here later?)  
- added a on mouse leave to the hamburger wallet  
- update route `/settings` to `/settings-network`  
- tweak message import / export modal content  
- add on hover effect to modal x  
- modify the wallet created styling  
- modify toast element
- modify and extract the trust wallet component
- adjust modal z factor
- add forgotten utxo compound to new menu - [Issue Raised](https://github.com/K-Kluster/Kasia/issues/90)
- adjust style of the wallets on home page. remove buttons, add trash icon, add pulse to migrate

### files added
- requireunlockedwallet – this is a guard for oneliner  
- rootlayout.tsx – root layout w/ outlet for app  
- ui.store.ts – store for menu settings  
- approutes.tsx – for routing  

### delete
- messaging-ghost.code-workspace – not in use  

### css
- take out a bunch of css from wallet guard  

### later to do (not part of this pr)
- extract the modals to the ui store (removing the context)  
- re-jig mobile layout of wallets  
- add network settings to use the layout

PR #92, opened 2025-06-28
Feat/mobile tweaks

HocusLocusTee:
Wait until [this](https://github.com/K-Kluster/Kasia/pull/89) is merged before merging.

## What?

- locks the view port on mobile, stops auto zooming etc. currently due to some elements being so small modern browsers will zoom for accessibility (if font is less than 16px for example). 
_I will address some of these in a subsequent PR once the current PRs are merged_
- mobile styling changes, primarily removes the margin space on mobile, making the app feel more 'native'. Side effects is wallet guard is now full-x (which is ok), but difficult to have control over without a further refactor.

**before:**
image

**after:**
image

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,11 +1,13 @@
-import React, { useCallback } from 'react'
+import React, { useCallback, useEffect } from 'react'
 import { useNetworkStore } from './store/network.store'
 import type { NetworkType } from './types/all'
 import { AppRoutes } from './AppRoutes';
+import { useIsMobile } from './utils/useIsMobile';
 
 const App: React.FC = () => {
   const networkStore = useNetworkStore();
   const connect = useNetworkStore((s) => s.connect);
+  const isMobile = useIsMobile();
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -15,6 +17,16 @@ const App: React.FC = () => {
     [connect, networkStore]
   );
 
+  useEffect(() => {
```
HocusLocusTee, line 20: @IzioDev wdyt?

IzioDev: Looks good for the logic, should we move this even before react is rendering?

Maybe as a direct `<script>` tag on the `index.html`?

HocusLocusTee: Up to you, the splash screen already *sort of* takes care of it (acts as a buffer).

I'd personally rather leave it here, but I'm good either way,..

IzioDev: Alright, let's leave it here

IzioDev:
LGTM, merging into the base branch and reviewing the base branch

PR #99, opened 2025-06-29
Further UI mods

HocusLocusTee:
## What?

a few ui mods:
- modify the new chat buttons and add Y space for address (so its no longer clipped)
- in the message pane, add a toast notification to when address is copied
- change styling of toast component
- space out nick name button for better fidelity 
- once payment is made - close popover in chat


src/components/ContactCard.tsx
```diff
@@ -180,21 +180,21 @@ export const ContactCard: FC<{
                 if (e.key === "Escape") handleNicknameCancel();
               }}
               autoFocus
-              placeholder="Enter nickname..."
+              placeholder={contact?.address}
```
IzioDev, line 183: Good call

IzioDev:
`588` lines of CSS until freedom. ❤️ 

```
$ wc **/*.css -l
  575 src/index.css
   13 src/variables.css
  588 total
```

LGTM.

PR #100, opened 2025-06-29
Feat add button plus misc

HocusLocusTee:
## What?

adds a new `Button` component

- restyles a lot of buttons on site with the new component
- standardises buttons through the above
- replace some errors with toast
- i fix bug where open menu state persists between logging in and out
- minor adjust to toast colors and its z
- misc fix to closing modals
- add a hover conditional to the feebuckets on desktop header

_note: this maybe not be our final colors etc for the button, but at least we can change it in one place now_

### DIFF
--- a/src/components/Common/Button.tsx
+++ b/src/components/Common/Button.tsx
@@ -1,0 +1,37 @@
+import clsx from "clsx";
+import { ButtonHTMLAttributes, FC } from "react";
+
+type ButtonProps = ButtonHTMLAttributes<HTMLButtonElement> & {
+  variant?: "primary" | "secondary";
+};
+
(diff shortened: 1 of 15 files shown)


src/components/Common/Button.tsx
```diff
@@ -0,0 +1,32 @@
...
+type ButtonProps = ButtonHTMLAttributes<HTMLButtonElement> & {
+  variant?: "primary" | "secondary";
+};
+
+export const Button: FC<ButtonProps> = ({
+  variant = "primary",
+  disabled,
+  className,
+  children,
+  ...props
```
IzioDev, line 13: awesome, wen loading state? wen icon? 😆

PR #106, opened 2025-06-30
Style/homescreen

HocusLocusTee:
## What?

major styling overhaul for wallet-flow aka the locked homescreen

changes:
- replace radio buttons with headless ui radio groups
- restyle input fields
- adjust spacing
- softened a bunch of colors
- adjusted useIsMobile hook due to bug
- take out css for tailwind

on mobile:
- made the home-screen 'full screen'

notes:
- aware of the repetition in the component adjustments, the whole wallet flow needs a refactor (there are bigger things to work on though) so we can deal with it then. this should make it easier.
- I ended up getting *very* hacky with the wrapper class styling, mainly because I don't want all the items at the top of the page. some styling that would *make sense* adds other undesirable effects.

## Testing
nest everything, do a full flow of: creating wallet > unlocking > check address > deleting > importing the same wallet > check address > migrate > delete > restore.

note: none of the logic has changed, but last thing I want is that there's a bug in something critical like expected behavior for wallet gen.

note-2: if its all good - merge, deploy to staging and see how it feels on a real wallet. adjust accordingly

### DIFF
--- a/src/components/Layout/TrustMessage.tsx
+++ b/src/components/Layout/TrustMessage.tsx
@@ -9,7 +9,7 @@ export const TrustMessage: FC = () => {
   return (
     <>
       {/* trust message section */}
-      <div className="mt-8 rounded-lg border border-[#70C7BA]/20 bg-gradient-to-br from-[#70C7BA]/10 to-[#70C7BA]/5 p-2">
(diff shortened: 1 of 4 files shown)


IzioDev:
That's massive @HocusLocusTee, you did a super great job on this, thanks! ❤️

PR #110, opened 2025-07-01
UI and network fix

HocusLocusTee:
## What?

ui bug fixes:
- incorrect label name
- tidy up create wallet seed area, remove css
- added a quick and dirty disable and tooltip on the unlock button, we could bring the network selector in here. But I've run outta time.

network connection bug fix:
bug: when changing networks (testnet / mainnet) the connection dot stays green, seems like we had a false isConnected state at the start of connection attempts (not initial, but subsequent)

### DIFF
--- a/src/containers/WalletFlow.css
+++ b/src/containers/WalletFlow.css
@@ -85,28 +85,6 @@
   opacity: 0.9;
 }
 
-.mnemonic-word-input {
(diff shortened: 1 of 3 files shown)


IzioDev:
nice catch about the bug!

it also occurred while configuring your own node url

PR #116, opened 2025-07-02
fix(offline): history retrieval

IzioDev:
* fix offline message retreival system, which still have flaws that should be documented as new issues
  * message history cannot be greater than 50 by design, could be fine to fetch latest, but eventually one should be able to tailor it's refinement by only relecting a sending address
  * could be done in async because it's heavy in term of boot time i've noticed (is it just me?)

⚠️ i had to add `ts-ignore`, this was the fastest way, but ideally we would lie to on another level to avoid errors in the future (function signature changes)


HocusLocusTee:
@IzioDev 
I re-based from staging and pushed. You didn't have the speed up commit in its history - so it seemed longer to start the message client.

PR #119, opened 2025-07-04
Feat/avatar hash

HocusLocusTee:
## What?

2 things:
- restyles the contacts pane so the contacts follow our general theme a little better (boarder, on hover, color etc)
- adds deterministic dot pattern for contacts when collapsed! 
  - this is a bit of a feature, when selected the dot pattern will animate by adding dots, pausing and reducing them. Also a soft blur around the edges
  - adding a nickname will show the dot pattern still but include the first letter of the nickname

### test it:
- checkout the branch
- test the first feature: open your wallet, try your mouse over the contacts when not collapsed
- second feature: collapse the contacts check out dot pattern, select a contact, add a nickname etc

### DIFF
--- a/src/components/ContactCard.tsx
+++ b/src/components/ContactCard.tsx
@@ -180,7 +180,7 @@ export const ContactCard: FC<{
   return (
     <div
       className={clsx(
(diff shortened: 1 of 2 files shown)


IzioDev:
LGTM, looks good.

Maybe could spend some time on tweaking in the future regarding colors, loop time, glow effect, but it's details

PR #124, opened 2025-07-04
Fix/improve message loading

AuzGhosty:
Non-Blocking Loading
- Historical messages load in the background without blocking the UI
- Users can start sending/receiving messages while older messages are still loading

404 Handling
- Previously: Would retry 404s up to 10 times with 2-second delays (causing 20+ seconds of waiting)
- Now: Immediately marks 404 transactions as "too old" and skips them
- Maintains a cache to avoid re-checking known old transactions


IzioDev:
We can close this one as the historical loading has been completely rewrote

Issue #109, opened 2025-06-30
Unknown Contacts


HocusLocusTee:
Sometimes when receiving a payment from a known and active contact - the address comes through as unknown:

![Image](https://github.com/user-attachments/assets/6a4785bf-183d-4e9b-ba54-15ca3a30a1f8)

![Image](https://github.com/user-attachments/assets/9800ec08-514b-4ea4-958a-f913dcb01f81)

HocusLocusTee:
Can confirm this still will happen if the API hasn't see the tx yet and all our requests to get the address fail.

HocusLocusTee:
Done.

PR #127, opened 2025-07-05
Fix/uiandpayment

HocusLocusTee:
## what?

style fixes:
allows crude line breaking on toast messages and payment messages
adds cursor pointers to payment buttons

in chat payment:
fixes back where max send tried to send all funds without taking into account the payload and fees

### DIFF
--- a/src/service/account-service.ts
+++ b/src/service/account-service.ts
@@ -1163,15 +1163,36 @@ export class AccountService extends EventEmitter<AccountServiceEvents> {
 
     console.log("Using destination address:", transaction.address.toString());
 
+    // check if the funds being sent are full balance
+    const isFullBalance = transaction.amount === this.context.balance?.mature;
+    console.log("Full balance check:", {
+      requestedAmount: transaction.amount.toString(),
+      matureBalance: this.context.balance?.mature.toString(),
+      isFullBalance,
+    });
+
+    // if thats the case, use destination as change address and ReceiverPays fees
+    const changeAddress = isFullBalance
+      ? new Address(transaction.address.toString())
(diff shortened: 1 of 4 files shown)


src/service/account-service.ts
```diff
@@ -1163,15 +1163,36 @@ export class AccountService extends EventEmitter<AccountServiceEvents> {
 
     console.log("Using destination address:", transaction.address.toString());
 
+    // check if the funds being sent are full balance
+    const isFullBalance = transaction.amount === this.context.balance?.mature;
```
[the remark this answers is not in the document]
HocusLocusTee: Estimating this now.

src/service/account-service.ts
```diff
@@ -1244,7 +1276,7 @@ export class AccountService extends EventEmitter<AccountServiceEvents> {
             );
             if (prevTx?.outputs && prevTx.outputs[prevOutputIndex]) {
               const output = prevTx.outputs[prevOutputIndex];
-              senderAddress = output.verboseData?.scriptPublicKeyAddress;
+              senderAddress = output.script_public_key_address;
```
HocusLocusTee, line 1279: See note on this in [discord](https://discord.com/channels/1359255794675089558/1360035142172676226/1391005000796340376).
This property doesn't exist, so we always fall back to the logic below.

But in balance emptying tx like 'send all', we are excluded from the fall back logic via outputs.length>1.

This change is more accurate and fixes that.

HocusLocusTee: Might solve this [issue](https://github.com/K-Kluster/Kasia/issues/109)

Issue #135, opened 2025-07-06
Start new chat using QR code


HocusLocusTee:
As a user, it would be useful if when starting a new chat, there's an option to use your camera to scan a QR code of their address.

PR #138, opened 2025-07-07
feat: qrscan + misc

HocusLocusTee:
## what
- adds `jsqr` for decoding qr codes (note: jsqr doesn't handle camera input, only decoding from images, its bundle size is comparatively good). 

- adds the `QrScanner` component for scanning and the likes

- enables qr code scanning in new chats and withdraw. 

note: unlike the other soft standard where we have a parent button that calls a component, this component includes its own button so it can hide if the useEffectt detects no camera. sorta an all in one type thing.

styling is bare minimum and I haven't tested on desktop.

also includes small changes: 
- adjusted width of the toast container for better display on mobile (well it fixes incomplete tailwind - from earlier me).

- added classnames as a prop to our common modal component for easier styling overrides.

## testing
needs https for camera access. used ngrok personally but use whatever works or test on staging.

### DIFF
--- a/src/components/Common/modal.tsx
+++ b/src/components/Common/modal.tsx
@@ -1,14 +1,19 @@
 import { FC, ReactNode } from "react";
 import { XMarkIcon } from "@heroicons/react/24/solid";
+import clsx from "clsx";
 
 // Basic modal component to standardise the look
 // Pass in your children - this just provides the bare minimum
-export const Modal: FC<{ onClose: () => void; children: ReactNode }> = ({
-  onClose,
-  children,
-}) => (
+export const Modal: FC<{
+  onClose: () => void;
+  children: ReactNode;
+  className?: string;
+}> = ({ onClose, children, className }) => (
   <div
-    className="modalFadeIn fixed inset-0 z-50 flex items-center justify-center bg-black/50"
+    className={clsx(
+      "fixed inset-0 z-50 flex items-center justify-center bg-black/50",
+      className
+    )}
     onClick={onClose}
   >
     <div
--- a/src/components/QrScanner.tsx
+++ b/src/components/QrScanner.tsx
@@ -1,0 +1,150 @@
+import React, { useRef, useEffect, useState } from "react";
+import jsQR from "jsqr";
+import { CameraIcon } from "@heroicons/react/24/solid";
+import { toast } from "../utils/toast";
+import clsx from "clsx";
+import { Modal } from "./Common/modal";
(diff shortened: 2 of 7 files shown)


HocusLocusTee:
Addresses [this](https://github.com/K-Kluster/Kasia/issues/135)

src/components/Common/modal.tsx
```diff
@@ -3,12 +3,13 @@ import { XMarkIcon } from "@heroicons/react/24/solid";
...
-  children,
-}) => (
+export const Modal: FC<{
+  onClose: () => void;
+  children: ReactNode;
+  className?: string;
+}> = ({ onClose, children, className = "" }) => (
   <div
-    className="fixed inset-0 z-50 flex items-center justify-center bg-black/50"
+    className={`fixed inset-0 z-50 flex items-center justify-center bg-black/50 ${className}`}
```
IzioDev, line 12: shouldn't we use `clsx`?

src/components/QrScanner.tsx
```diff
@@ -0,0 +1,150 @@
...
+    }
+    let localStream: MediaStream | null = null;
+    let cancelled = false;
+
+    const waitForVideoAndStartCamera = async () => {
+      let attempts = 0;
+      while (!videoRef.current && attempts < 20) {
+        await new Promise((resolve) => setTimeout(resolve, 50));
+        attempts++;
+      }
```
IzioDev, line 43: Why do we need to try multiple times?

src/components/Common/modal.tsx
HocusLocusTee: True, I was just thinking to 'add these classes' and mentally reserve clsx for conditionals.

src/components/QrScanner.tsx
HocusLocusTee: I had an early timing issue with trying to boot the camera too quickly (at least in my ghetto https testing environment to phone).  So I figured we may as well just give it a few tries.

src/components/Common/modal.tsx
HocusLocusTee: Resolved.

IzioDev: I personally am ok with this way of doing if you feel it's more natural / simple.
Should we all agree that `clsx` is only mandatory when using conditionals styles?

HocusLocusTee: No you're right, let's use clsx in all cases.

Issue #137, opened 2025-07-06
KNS or Nicknames not being displayed to header in message section


HocusLocusTee:
In the Message Section we display the active address at the top. 
We should show the KNS or nick if its set, but allow copying of the underlying address till.

PR #141, opened 2025-07-08
feat: propagate nickname to message section header

HocusLocusTee:
## What?
Nicknames can now be displayed in he message section header.

image

### DIFF
--- a/src/components/KaspaAddress.tsx
+++ b/src/components/KaspaAddress.tsx
@@ -2,33 +2,40 @@ import { FC, useMemo, useState, useEffect, useRef } from "react";
 import { Square2StackIcon } from "@heroicons/react/24/outline";
 import { toast } from "../utils/toast";
 import { useIsMobile } from "../utils/useIsMobile";
+import { useMessagingStore } from "../store/messaging.store";
 
 interface KaspaAddressProps {
   address: string | { toString: () => string };
 }
 
 export const KaspaAddress: FC<KaspaAddressProps> = ({ address }) => {
+  const NICKNAME_SHORT_LENGTH = 20;
   const [isFullAddress, setIsFullAddress] = useState(false);
   const addressRef = useRef<HTMLSpanElement>(null);
-  const isMobile = useIsMobile();
+
+  const addressStr = typeof address === "string" ? address : address.toString();
+  const contact = useMessagingStore((s) =>
+    s.contacts.find((c) => c.address === addressStr)
+  );
+  const nickname =
+    contact && typeof contact.nickname === "string" && contact.nickname.trim()
+      ? contact.nickname.trim()
+      : "";
+
+  const displayString = nickname || addressStr;
+  const isNickname = !!nickname;
 
   const [firstSubPart, secondSubPart] = useMemo(() => {
-    const asString = typeof address === "string" ? address : address.toString();
-    if (asString.length < 10) {
-      return [asString, ""];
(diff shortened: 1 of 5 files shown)


src/components/KaspaAddress.tsx
```diff
@@ -10,25 +11,39 @@ interface KaspaAddressProps {
...
-    const indexOfColon = asString.indexOf(":");
+  const [firstSubPart, secondSubPart] = useMemo(() => {
+    if (isNickname) {
+      if (displayString.length <= 20) {
+        return [displayString, ""];
+      }
 
-    // keep the prefix, can be either kaspa or kaspatest
-    const prefix = asString.slice(0, indexOfColon);
+      return [displayString.slice(0, 15), ""];
```
IzioDev, line 33: If less than 20 (19) keep this length, else reduce to 15?
should we add some consistency here, like if less than `x` then `x` else `y`?

src/components/KaspaAddress.tsx
```diff
@@ -2,6 +2,7 @@ import { FC, useMemo, useState, useEffect, useRef } from "react";
 import { Square2StackIcon } from "@heroicons/react/24/outline";
 import { toast } from "../utils/toast";
 import { useIsMobile } from "../utils/useIsMobile";
+import { useMessagingStore } from "../store/messaging.store";
 
 interface KaspaAddressProps {
   address: string | { toString: () => string };
```
IzioDev, line 8: I feel like we should add explicit optional props:
```ts
{
  address?: string | { toString: () => string };
  nickname?: string;
}
```

It needs to have at least one to operate, if both, nickname take precedence?
What do you think?

src/containers/ContactSection.tsx
```diff
@@ -36,6 +37,9 @@ export const ContactSection: FC<ContactSectionProps> = ({
   const collapsedW = "w-14";
   const isMobile = useIsMobile();
   const toggleSettings = useUiStore((s) => s.toggleSettings);
+  const setSelectedContactAddress = useMessagingStore(
+    (s) => s.setSelectedContactAddress
+  );
```
IzioDev, line 43: Don't we have openedRecipient?

Maybe we should define `openedRecipient` as:
```ts
{
  address: Address, // or string depending on usage
  nickname?: string,
  // ... other relevant props, if any
}
```
What do you think?

src/components/KaspaAddress.tsx
HocusLocusTee: See new implementation.

src/containers/ContactSection.tsx
HocusLocusTee: I think I stayed up too late last night. I have no idea why i did what I did haha. I wasn't even using selected contact? Anyway, review latest version please.

src/components/NewChatForm.tsx
```diff
@@ -278,7 +278,7 @@ export const NewChatForm: React.FC<NewChatFormProps> = ({ onClose }) => {
       // Initiate handshake with custom amount
 
       await messageStore.initiateHandshake(knsRecipientAddress, amountSompi);
-      messageStore.setOpenedRecipient(recipientInputValue);
+      messageStore.setOpenedRecipient(knsRecipientAddress);
```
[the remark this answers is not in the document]
HocusLocusTee: [Issue](https://github.com/K-Kluster/Kasia/issues/137)

Issue #150, opened 2025-07-11
Bug: Message sending process should be awaiting instead of allowing multiple sends


IzioDev:
> Whenever I send a message to my friends, it doesn’t get delivered the first time. I have to tap send again, and suddenly, the same message gets sent twice. This happens quite often, and it becomes confusing because they receive duplicate messages from me even though I only typed it once.

[source](https://discord.com/channels/1359255794675089558/1359255795648172245/1393170391941382246)

Proposed solution:
> adding a loading state to the send button, and disallow logic to be executed if a message is in the process of being sent (simple workaround, later on we'll add better handling of this)

HocusLocusTee:
"It doesn't get delivered", is interesting, as in they don't see it in chat (this is not delivery, this is local storage) or the receiver doesn't see it?
Ideally we _dont_ block the message send. A very small timeout is probably ok, but I'd rather work on a robust queuing system. 
On iMessage (in times of bad connection) the message propagates to the chat but you don't get the 'delivered' stamp.
Signal also has a status icon thing.

IzioDev:
If one is interested into implementing the end-goal straight away, i'm all in

cafalchio:
@IzioDev are you already doing it? I would like to have a try, unless you are going for the full solution

HocusLocusTee:
Once other code is merged I think we close this and open more specific issues: Such as [this](https://github.com/K-Kluster/Kasia/issues/166)

PR #161, opened 2025-07-20
Added status pending | send | failed  in Message to address delays

cafalchio:
It fix the git issue #150 

image


HocusLocusTee:
My initial initial (obviously non mandatory) visual feedback:
- switch text to an icon
- keep it the same text colour as the message (this helps with themes and just a consistent style overall)

cafalchio:
I will create a new one on staging branch

PR #162, opened 2025-07-21
Feature/pending failed message icons

cafalchio:
This commit is a fix to issue #150 and also adding icons to show pending and failed messages.

image

### DIFF
--- a/src/components/MessageDisplay.tsx
+++ b/src/components/MessageDisplay.tsx
@@ -7,7 +7,7 @@ import { CipherHelper } from "../utils/cipher-helper";
 import { useMessagingStore } from "../store/messaging.store";
 import { HandshakeResponse } from "./HandshakeResponse";
 import { KasIcon } from "./icons/KasCoin";
-import { Paperclip, Tickets } from "lucide-react";
+import { Paperclip, Tickets, Loader2, X } from "lucide-react";
 import clsx from "clsx";
 import { parseMessageForDisplay } from "../utils/message-format";
 import { PROTOCOL_PREFIX, PAYMENT_PREFIX } from "../config/protocol";
@@ -157,28 +157,6 @@ export const MessageDisplay: FC<MessageDisplayProps> = ({
       );
     }
     return null; // Don't show amount for any messages
-  };
-
-  // Render status for failed or pending
-  const formatStatus = () => {
-    if (status === "failed") {
-      return (
-        <div className="w-full">
-          <div className="message-fee text-right text-xs text-red-500">
-            {status}
-          </div>
-        </div>
-      );
-    }
-    if (status === "pending") {
-      return (
-        <div className="w-full">
-          <div className="message-fee text-accent-yellow text-right text-xs">
-            {status}
-          </div>
-        </div>
-      );
-    }
   };
 
   // Render payment message content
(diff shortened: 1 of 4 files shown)


src/components/MessageDisplay.tsx
```diff
@@ -489,16 +490,20 @@ export const MessageDisplay: FC<MessageDisplayProps> = ({
...
+              href={getExplorerUrl(transactionId)}
+              target="_blank"
+              rel="noopener noreferrer"
+              className="text-xs opacity-80 transition-opacity hover:opacity-100"
+            >
+              <Tickets className="size-5" />
+            </a>
+          )}
+          {status === "pending" && <Loader2 className="size-5 animate-spin" />}
+          {status === "failed" && <X className="size-5 text-red-500" />}
```
HocusLocusTee: can we use ` --accent-red` from `index.css` instead?

HocusLocusTee:
I think this helps towards [Issue ](https://github.com/K-Kluster/Kasia/issues/150). But we can still send / spam messages (leading to failures) and its not really fix. 

Further more to this, if you send a message before the fee estimation is finished. You can get toast errors.

HocusLocusTee:
Also, in some sates we get messages that have none of the status symbols:
image

See my comment below, it might address this.

src/containers/SendMessageForm.tsx
```diff
@@ -231,7 +233,26 @@ export const SendMessageForm: FC<SendMessageFormProps> = ({ onExpand }) => {
         // Not a file message, use message as is
       }
 
-      let txId: string;
+      let txId: string = "";
+
+      // Create the message object for storage
+      pendingMessage = {
+        transactionId: "",
```
HocusLocusTee: If you spam messages, we get issues saying there is duplicate keys in the render (which is duplicate empty string keys).

Maybe we can use a uuid here or just do: 
>pendingMessage = { transactionId: `pending-${Date.now()}-${Math.random()}`

 for pending tx. 
That way we wont even have a null id and try to render it... It should solve us from not ever having an icon and we'll never throw the warning.

cafalchio: I will spend more time on it, because not my if txId == "" will be always false.

cafalchio: I will spend a.bit more time on it.

cafalchio: This last commit fix the messages when spamming.

HocusLocusTee:
Another discussion point:
If I try to send a tx and fee estimation is not complete. It just sends the tx and fails it.
Leaving the message text still in the input.

I think this is going to be a v0.4.0 addition.
I don't want to rush this through just because we have a deadline with v0.3.0

image

cafalchio:
Do you think we need to wait for estimated fee before letting the user to click send?

HocusLocusTee:
> Do you think we need to wait for estimated fee before letting the user to click send?
Thats how it currently is on staging.

I think there is a better solution somewhere. 
Even if we don't wait.. we shouldn't do the following: fail, not retry sending and leave a failed message bubble.

cafalchio:
100% agree.

HocusLocusTee:
Closing this due to being many revisions behind and https://github.com/K-Kluster/Kasia/pull/278 now existing

PR #165, opened 2025-07-23
Fix/style+i omessages

HocusLocusTee:
### style:
- fix settings on mobile: icon width, top border
- message bubble - slightly soften sharp corner
- primary button had lost its hover

### message import export
- fixed my previous mistake. now working with new encryption storage
- adjusted file name to include some wallet char
- made sure nicknames were applied straight away on import


IzioDev:
LGTM, will cause a few conflicts with current idb branch, but this is needed for v0.3.0 anyway 👍

Issue #169, opened 2025-07-26
Extract payload (strings) to `config/protocol.ts``


HocusLocusTee:
I've already extracted the hex strings. Maybe lets take them to the above and make them a simple object.

PR #170, opened 2025-07-27
Refactor/protocol const /1

HocusLocusTee:
## what?
this just refactors and centralizes the protocol prefix / headers.
aware its due to change with dcom, but this gives it some structure.

for review ease, i just modify `src/config/protocol.ts` and then propagate its use throughout the app.

[this card](https://github.com/K-Kluster/Kasia/issues/169)

note: there is some double handling, like calling the string version of the protocol and then calling .hex. we can sort this out after rebase is finished.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "kasia",
-  "version": "0.2.1",
+  "version": "0.3.0",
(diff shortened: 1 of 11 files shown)


IzioDev:
LGTM

PR #171, opened 2025-07-27
Refactor/parsemessage /2

HocusLocusTee:
## what?
extracts parsemessage function into a new utility
and spreads its use

note: we lost a bunch of use from this extra file, we can deal with it post rebase.


IzioDev:
LGTM

PR #172, opened 2025-07-27
feat:add featureflip system and functionality /3

HocusLocusTee:
## what?
this adds a user feature flip system to the settings modal menu
currently stores status in local storage. potentially will migrate to indexeddb later too.

note 1.a: there's nothing in the store atm, there are 2 'example' flags in the `src/store/featureflag.store.ts` store.
note1.b: you *will* need to comment these out to test this pr.

2 open questions:
also not sure if we should have a flag set *per wallet* or just generally for the account?
not sure if I should move this from extras tab to the account settings?

^potentially it could be unified with theme under `settings-{wallet}` but ill wait for indexeddb for that

### DIFF
--- a/src/store/featureflag.store.ts
+++ b/src/store/featureflag.store.ts
@@ -38,27 +38,28 @@ const useFeatureFlagsStore = create<{
   flags: FeatureFlags;
   flips: FeatureFlip[];
   setFlag: (key: keyof FeatureFlags, value: boolean) => void;
-  loadFlags: () => void;
-}>((set, get) => ({
-  flags: defaultFlags,
-  flips: featureFlips,
+}>((set, get) => {
+  // hydrate features here, else take default value
+  let initialFlags = defaultFlags;
+  try {
+    const stored = JSON.parse(
+      localStorage.getItem("kasia-feature-flags") || "{}"
+    );
+    initialFlags = { ...defaultFlags, ...stored };
+  } catch {
+    console.error("Invalid flags in localStorage");
+  }
 
-  setFlag: (key, value) => {
-    const updated = { ...get().flags, [key]: value };
-    set({ flags: updated });
-    localStorage.setItem("kasia-feature-flags", JSON.stringify(updated));
-  },
+  return {
+    flags: initialFlags,
+    flips: featureFlips,
 
-  loadFlags: () => {
-    try {
-      const stored = JSON.parse(
-        localStorage.getItem("kasia-feature-flags") || "{}"
-      );
-      set({ flags: { ...defaultFlags, ...stored } });
-    } catch {
-      console.error("Invalid flags in localStorage");
-    }
-  },
-}));
+    setFlag: (key, value) => {
+      const updated = { ...get().flags, [key]: value };
+      set({ flags: updated });
+      localStorage.setItem("kasia-feature-flags", JSON.stringify(updated));
+    },
+  };
+});
 
 export { useFeatureFlagsStore };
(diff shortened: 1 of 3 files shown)


IzioDev:
> currently stores status in local storage. potentially will migrate to indexeddb later too

No problem of keeping this information in the local storage, it's probably not an infinitely grow-able data store :)

> not sure if we should have a flag set per wallet or just generally for the account

It's probably a user-based setting more than a wallet-based setting. could be arguable on specific settings in the future

> not sure if I should move this from extras tab to the account settings?

let's start the most straightforward way, and if we feel like it starts to be complicated, we'll change our position

What do you think?

src/store/featureflag.store.ts
```diff
@@ -0,0 +1,64 @@
...
+  //   desc: "This enables external game frame content. NOT IMPLEMENTED - JUST FOR FLAG DEMO.",
+  //   icon: Dices,
+  // },
+  // {
+  //   id: "moreitems",
+  //   label: "More Items",
+  //   desc: "This would be something else!",
+  //   icon: Dices,
+  //},
+];
```
IzioDev: What do you think to tie flag with feature like this?
```ts
export enum FeatureFlags {
  GAME_FRAME = 'gameframe',
  MORE_ITEM = 'more_item',
}

export type FeatureFlagsTable = Record<FeatureFlags, boolean>;

const defaultFeatureFlagsTable: FeatureFlagsTable = {
  [FeatureFlags.GAME_FRAME]: false,
  [FeatureFlags.MORE_ITEM]: false,
};

export interface FeatureDescription {
  label: string;
  desc: string;
  icon: string;
}
export type FeatureFlips = Record<FeatureFlags, FeatureDescription>;

const featureFlips: FeatureFlips = {
  [FeatureFlags.GAME_FRAME]: {
    label: 'Game frame',
    desc: 'Game frame desc',
    icon: '...',
  },
  [FeatureFlags.MORE_ITEM]: {
    label: 'MORE_ITEM',
    desc: 'MORE_ITEM',
    icon: '...',
  },
};
```

IzioDev: This forces to define all feature flips (type error if not), without the need to re-refine `id`.

Using an enum to be able to add type doc along it, and potentially easier access `FeatureFlags.xxx`

src/store/featureflag.store.ts
```diff
@@ -0,0 +1,64 @@
...
+  loadFlags: () => {
+    try {
+      const stored = JSON.parse(
+        localStorage.getItem("kasia-feature-flags") || "{}"
+      );
+      set({ flags: { ...defaultFlags, ...stored } });
+    } catch {
+      console.error("Invalid flags in localStorage");
+    }
+  },
```
IzioDev: Since this is an hydration that isn't dependent of the state of the application, we could even load the flags like:

```ts
create<{...}>((set, get) => {
  // hydrate features here, else take default value


  return (
    {
      // methods and state default here
    }
  )
})
```

This would remove the need for calling this hydration manually

HocusLocusTee:
> > currently stores status in local storage. potentially will migrate to indexeddb later too
> 
> 
> 
> No problem of keeping this information in the local storage, it's probably not an infinitely grow-able data store :)
> 
> 
> 
> > not sure if we should have a flag set per wallet or just generally for the account
> 
> 
> 
> It's probably a user-based setting more than a wallet-based setting. could be arguable on specific settings in the future
> 
> 
> 
> > not sure if I should move this from extras tab to the account settings?
> 
> 
> 
> let's start the most straightforward way, and if we feel like it starts to be complicated, we'll change our position
> 
> 
> 
> What do you think?

Ack on all this.

Dbstore was mainly to align with any encryption / decryption in the future. But we can just do the saw to the local storage (small amount of data).

PR #176, opened 2025-07-29
Fix/migration issues

HocusLocusTee:
migration fix update

this should be fixed now.

**what i tested:**
- created multiple wallets with messages on the old `master` branch
- checked out the new branch and logged into a wallet → messages migrated
- logged into a second wallet → messages migrated
- confirmed that once all wallets are migrated, old local storage is deleted
- on the new branch: verified support for importing:
  - encrypted wallets
  - unencrypted wallets
  - previously encrypted formats

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "kasia",
-  "version": "0.2.1",
+  "version": "0.3.0",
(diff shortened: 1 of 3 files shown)


IzioDev:
Awesome, thank you very much @HocusLocusTee !

PR #179, opened 2025-07-29
Feat/react compiler/5

HocusLocusTee:
## what?
part 1:
introduces react compiler

removes modaldisplay and introduces modalhost component < which is state compatible with the compiler
fixes numerous hook violations
changes splash.tsx out for splash.ts

[Related](https://github.com/K-Kluster/Kasia/issues/177)

part 2:
~remove all - now redundant - manual optimizations such as `useMemo` and `useCallback`~

[Solves this](https://github.com/K-Kluster/Kasia/issues/177?issue=K-Kluster%7CKasia%7C178)

## testing
whether its now or later, we do need to test full flows in-case there is any ui state bugs

**update:
rebased on staging, removed all optimizations. will do later**


vite.config.ts
```diff
@@ -6,7 +6,18 @@ import { VitePWA } from "vite-plugin-pwa";
...
-    react(),
+    react({
+      babel: {
+        plugins: [
+          [
+            "babel-plugin-react-compiler",
+            {
+              panicThreshold: "none",
+            },
+          ],
```
IzioDev, line 17: This should be temporary, do you confirm?

IzioDev:
Besides the question I had regarding disabling rules of react check, LGTM

vite.config.ts
HocusLocusTee: Nay, [docs show](https://react.dev/reference/react-compiler/panicThreshold#caveats):
> Caveats
Production builds should always use 'none'

PR #181, opened 2025-07-29
fix(migration): segregate legacy messages per address

IzioDev:
`migrateToPerAddressStorage` now takes the wallet address argument that is used to filter `legacyMessages`. Previously we considered that legacyMessages would contains only legacy message on the loaded wallet, now we take that consideration and filter accordingly.


HocusLocusTee:
this is superseded by [this](https://github.com/K-Kluster/Kasia/pull/183)

PR #183, opened 2025-07-31
fix: load and import messages

HocusLocusTee:
## what?

lgtm


IzioDev:
also tested on my end, working fine, thanks again!

PR #188, opened 2025-08-02
Refactor/locked wallet /7

HocusLocusTee:
## what?
refactors locked wallet breaking it up into some sub components, and adding `WalletLockedFlowContainer`
fixes styling bug (theme consistency) in the splash page
changes wallet flow to inline errors (no toast while locked) < not sure about this

**update:
rebased on staging**


src/components/WalletLockedFlow/ErrorMessage.tsx
```diff
@@ -0,0 +1,17 @@
+type ErrorMessageProps = {
+  message: string;
+  className?: string;
+};
+
+export const ErrorMessage = ({
```
IzioDev: Maybe we should prefix the component name by its "scope", like `WalletFlowErrorMessage` or something similar, unless that is supposed to be a more generic component, and should be move on the general scope

The primary reason for this if when I want to use `ErrorMessage`, within the auto-complete I will most likely see multiple of them

What do you think?

IzioDev (commented):
Looks great, step by step we'll get there! 💪

src/components/WalletLockedFlow/ErrorMessage.tsx
HocusLocusTee: I agree with refined naming (it's only meant to be used while locked).
What are you thoughts on inline error messages while locked, but only toast notifications while logged in?

The sole reason I did this was because it made the refactoring v easy. But, since locked wallet fields are all 'primary' inputs, I stuck to it.

IzioDev: I see no major UX issue while using toast only for error while unlocked.

My only concern would be: as a phone user, if i enter the wrong password, can i still understand what's going on, do I see the error message properly?

> But, since locked wallet fields are all 'primary' inputs, I stuck to it.

I'm not sure I understand this properly. what are locked wallet fields? to be it's only password isn't it?

IzioDev: Ok got it, it also include create/import flows

IzioDev:
LGTM

PR #191, opened 2025-08-04
refactor: message input in chat /8

HocusLocusTee:
## what?
this refactors the send message form 
deleting it, adding a store and trying to handle _everything_ just a little bit better.

hopefully paves the way for if and when we get rich text editor like tiptap

it loosely sets a soft standard for following store / hook interactions, discussed in discord [around here](https://discord.com/channels/1359255794675089558/1360035142172676226/1402177141793099786)

>1 hook for 1 store, and for complex case 1 hook for n hooks - n store? creating a "manager hook" on top of the others?

_store use for _pure_ state updates are fine, but if there are side effects, then we go through a hook._

**rebased
originally I had deleted `SendMessageForm` but the indexeddb has some changes to it (that I will need to migrate to the new files). I will do this once we are at parity with staging.**

### DIFF
--- a/src/hooks/MessageComposer/useFeeEstimate.ts
+++ b/src/hooks/MessageComposer/useFeeEstimate.ts
@@ -4,8 +4,12 @@ import { useWalletStore } from "../../store/wallet.store";
 import { Address } from "kaspa-wasm";
 
 export const useFeeEstimate = (recipient?: string) => {
-  const { draft, priority, sendState, setFeeState } = useComposerStore();
+  const { priority, sendState, setFeeState } = useComposerStore();
   const walletStore = useWalletStore();
+
+  const draft = useComposerStore((s) =>
+    recipient ? s.drafts[recipient] || "" : ""
+  );
 
   useEffect(() => {
     if (
--- a/src/hooks/MessageComposer/useFeeEstimate.ts
+++ b/src/hooks/MessageComposer/useFeeEstimate.ts
@@ -1,0 +1,48 @@
+import { useEffect } from "react";
+import { useComposerStore } from "../../store/message-composer.store";
+import { useWalletStore } from "../../store/wallet.store";
+import { Address } from "kaspa-wasm";
+
+export const useFeeEstimate = (recipient?: string) => {
+  const { draft, priority, sendState, setFeeState } = useComposerStore();
+  const walletStore = useWalletStore();
+
+  useEffect(() => {
+    if (
+      !recipient ||
+      !draft ||
+      !walletStore.unlockedWallet ||
+      sendState.status === "loading"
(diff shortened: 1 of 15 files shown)


src/hooks/MessageComposer/useFeeEstimate.ts
```diff
@@ -0,0 +1,48 @@
...
+    const timeoutId = setTimeout(() => {
+      walletStore
+        .estimateSendMessageFees(draft, new Address(recipient), priority)
+        .then((estimate) => {
+          const fee = Number(estimate.fees) / 100_000_000;
+          setFeeState({ status: "idle", value: fee });
+        })
+        .catch((error) => {
+          setFeeState({ status: "error", error: error as Error });
+        });
```
IzioDev: In an ideal setup what scope/domain would you attribute to `estimateSendMessageFees` (currently `wallet`)

IzioDev: There is a very small chance / time window in which the timeout has been cleared (component unmounting) but the `estimateSendMessageFees` promise was still being resolved which could lead to a race condition between mutators of `composerStore` (`setFeeState`)

I also understand the reason why we'd like to segregate actual fee calculation from composer "ui" state/store

src/hooks/MessageComposer/useAttachments.ts
```diff
@@ -0,0 +1,41 @@
+import { useComposerStore } from "../../store/message-composer.store";
+import { prepareFileForUpload } from "../../service/upload-file-service";
+import { MAX_PAYLOAD_SIZE } from "../../config/constants";
+import { toast } from "../../utils/toast-helper";
+
+export const useAttachments = () => {
```
IzioDev: proposal: rename `useAttachments` to `useAddAttachment` because this is the only exposed operation

general thought: we created an abstraction layer on top of `composerStore` and this layer should be the only entry-point for components wanting to add/process a file to an attachment. In this context, should we also consider renaming to `useAdaptFileToComposerAttachment` (or similar)?

This way we make it clear this hook is generic way of adding an attachment, while we still have the capability to set it "raw" via the store?

IzioDev (commented):
Pretty solid! 💪

src/hooks/MessageComposer/useFeeEstimate.ts
HocusLocusTee: Not message composer, as getting fees are also needed for payments / withdrawals / etc.

HocusLocusTee: addressed.

IzioDev:
Let me know if I can help to bring indexdb changes

PR #211, opened 2025-08-13
Handshakev2/mods

HocusLocusTee:
# what?
some mods to the handshake pr

remove non-contextual tx submission logic
- we can only ever send contextual messages from the message composer

remove costly warn modal 
- not needed, see above

remove indexer fetching if there is no alias. 
-this prevents a user from seeing messages if they have not accepted the handshake. although its possible to read them, its in best interest to protect people from spam or any wanted messages.

remove payment message if no alias
- this was a sidestep around the above logic, we now do not show the added messages (but we still show the funds landing) if handshake is not accepted.

after handshake accept, we poll indexer for historical messages
- due to the above

+ some misc refactors and renaming

### DIFF
--- a/src/components/Layout/ModalHost.tsx
+++ b/src/components/Layout/ModalHost.tsx
@@ -111,43 +111,6 @@ export const ModalHost = () => {
         </Modal>
       )}
 
-      {/* Warning Costly Send Message Modal */}
-      {modals["warn-costy-send-message"] && (
-        <Modal onClose={() => closeModal("warn-costy-send-message")}>
-          <div className="flex flex-col items-center justify-center gap-8">
-            <h2 className="text-lg text-yellow-400">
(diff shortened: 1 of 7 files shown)


IzioDev:
LGTM

Small question: we started to introduce indexer transaction within the account service, but it's still has the explorer dependency for live event (online fetching when receiving a message). Do we added that in anticipation of explorer dependency being completely removed? (near future PR)

HocusLocusTee:
> LGTM
> 
> Small question: we started to introduce indexer transaction within the account service, but it's still has the explorer dependency for live event (online fetching when receiving a message). Do we added that in anticipation of explorer dependency being completely removed? (near future PR)

Ahh that was an _assumption_ by me. If its still Explorer I will revert terminology before I merge.

HocusLocusTee:
Resolved, so I am merging.

PR #212, opened 2025-08-14
Handshakev2/mods2 fees and conversation status

HocusLocusTee:
## what?

splits pending state to sent-pending and receive-pending
blocks receiver of handshake to response if they dont accept

### DIFF
--- a/src/service/conversation-manager-service.ts
+++ b/src/service/conversation-manager-service.ts
@@ -125,7 +125,7 @@ export class ConversationManagerService {
         // Keep the first alias - reuse existing pending conversation
         if (
           conversationAndContact &&
-          conversationAndContact.conversation.status === "pending"
+          conversationAndContact.conversation.status === "sent-pending"
         ) {
           // Create handshake payload with the existing alias (keeps first alias)
           const handshakePayload: HandshakePayload = {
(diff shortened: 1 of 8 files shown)


src/service/conversation-manager-service.ts
```diff
@@ -203,40 +203,37 @@ export class ConversationManagerService {
...
 
       if (existingConversationAndContactByAddress) {
-        // ------- this is a replay of a message we already handled -------
-        // keep the guard so we don't downgrade on refresh
-        if (payload.isResponse) {
-          // Promote the existing pending conversation to active *in-place* so any listeners that hold the original object see the change immediately.
-          (
-            existingConversationAndContactByAddress.conversation as unknown as ActiveConversation
-          ).status = "active";
+        const conv = existingConversationAndContactByAddress.conversation;
```
HocusLocusTee, line 206: @IzioDev 
Pay attention to this block, I was a bit confused about the comments and the code you wrote. In hindsight I can assume some intention, but Ill leave it up to you

src/store/messaging.store.ts
```diff
@@ -237,7 +237,13 @@ export const useMessagingStore = create<MessagingState>((set, g) => {
...
+        set((s) => ({
+          oneOnOneConversations: s.oneOnOneConversations.map((oooc) =>
+            oooc.conversation.id === conversation.id
+              ? { ...oooc, conversation }
+              : oooc
+          ),
+        }));
       },
       onHandshakeExpired: (conversation, contact) => {
         console.log("Handshake expired:", conversation);
```
IzioDev, line 249: I believe all of these events operations on the conversation manager should be removed, and should be handled by the caller for better clarity, what do you think?

IzioDev: In fact, I wanted to remove all of that in an upcoming PR (still not done), but I think we can do that in this one if you feel it's ok

src/store/repository/conversation.repository.ts
```diff
@@ -1,7 +1,11 @@
 import { decryptXChaCha20Poly1305, encryptXChaCha20Poly1305 } from "kaspa-wasm";
 import { DBNotFoundException, KasiaDB } from "./db";
 
-export type ConversationStatus = "pending" | "active" | "rejected";
+export type ConversationStatus =
+  | "sent-pending"
+  | "receive-pending"
+  | "active"
+  | "rejected";
```
IzioDev, line 8: note: since we still don't have a deployed version of Index DB live, we don't have to migrate existing data

src/service/conversation-manager-service.ts
```diff
@@ -634,23 +635,60 @@ export class ConversationManagerService {
...
-      });
+          alias: conversation.myAlias,
+          address: contact.kaspaAddress,
+        })
+      );
 
     return monitored;
   }
 
+  public getConversationStatusByAddress(address: string): {
```
IzioDev, line 668: Should we name it `getConversationLightInfoByAddress`? 

Are these minimal information considered "standard" in a way we will not have multiple of these different "light info" method

src/service/conversation-manager-service.ts
```diff
@@ -340,7 +341,7 @@ export class ConversationManagerService {
     contact: Contact;
   }[] {
     return Array.from(this.conversationWithContactByConversationId.values())
-      .filter(({ conversation }) => conversation.status === "pending")
+      .filter(({ conversation }) => conversation.status === "receive-pending")
       .map(({ conversation, contact }) => ({
         conversation: conversation as PendingConversation,
         contact,
```
IzioDev, line 347: should we rename `getReceivedPendingConversationWithContact` instead to better reflect the applied filter?

IzioDev (commented):
I'm wondering the need for `send-pending`, once I have initiated a conversation with a recipient, I should not have different app behavior than if it was active, right?

If my comment makes sense, maybe we can use `receive-pending` only.

Before merging (or as part of a future PR), we probably should modify the migration logic, so that it changes the state to either `active` or `receive-pending` depending on who initiated the conversation (if my previous comment is accepted)

HocusLocusTee:
> I'm wondering the need for `send-pending`, once I have initiated a conversation with a recipient, I should not have different app behavior than if it was active, right?
> 
> If my comment makes sense, maybe we can use `receive-pending` only.
> 
> Before merging (or as part of a future PR), we probably should modify the migration logic, so that it changes the state to either `active` or `receive-pending` depending on who initiated the conversation (if my previous comment is accepted)

The need has come up after looking at fee estimation. 
If there is an active convo, we assume 0 outputs. But now that handshake behavior has changed, really all ''messages' have zero outputs.

So either we just include ''pending'' in the conditional for an activeConvo and find another way to block the scenario where bob has received but not accepted handshake.

Or we split it like this.

HocusLocusTee:
closing in favor for simpler approach

Issue #201, opened 2025-08-11
Increase Max Payload Size


HocusLocusTee:
Set payload size to an optimistic maximum (say 20kb)<br>Rework image upload to allow setting of sending lower quality (lower fee)<br><br>*Not replacing further research and actual calculation of proper payload size*

PR #217, opened 2025-08-16
Feat/base64 larger payloads

HocusLocusTee:
## what?

for standard comm messages, this branch now encodes our hex to unit8array before submitting to generator. this saves a reasonable amount of space.

this has only been applied to contextual `comm` messages, mainly due to conversation manager being removed in the near future (and that's where we would need to do the change). additionally, we don't hit size limits on payments or handshakes, so its safe for now.

when listening for events, we now try to convert from base64 > hex and fall back to hex (for legacy stuff and payments / handshakes etc).

also: 
- adjusted the max MB limit for files and images.
- increased image quality etc.
- added some char limits to chat (18k char limit)

[solves](https://github.com/K-Kluster/Kasia/issues/201)
## how much more can we stuff in a tx?
well here's the thing, the tx size will be the same, theres no *increase* in what we are submitting, we were already maxing these out due to the required bloat of encryption, and hex conversion (this is the important bit). we now just lower the space the hex takes up.

### some testing

trying to post tx with the follow file sizes below:

#### old branch
- 7 KB file → 21,103 mass  
- 8 KB file → 23,831 mass  
- 9 KB → 26,559 mass → **failed (storage mass > 100k in final tx)**

#### this branch
- 10 KB → 21,615 mass  
- 11 KB → 23,435 mass  
- 13 KB → 27,075 mass  
-  14 KB → 28,895 mass → **failed (storage mass > 100k in final tx)**

### file summary
- **old branch**: fails already at 9 KB raw.  
- **new branch**: survives up to 13 KB raw, fails at 14 KB.  
- **conclusion**: new branch packs **larger raw files** before hitting the 100k storage mass ceiling. 

question: "uhhhhhhh that's still not very big, shouldn't we be able to send larger files?"
answer: true, we lose a lot of space with files and images with the 'metadata' and type we sorta wrap it up in. but this pr doesnt address that

### if we look at raw message chat length:
the old branch could handle 13k characters
this new branch can now handle  18k characters

**what can be done in the future?**
well this was an improvement based on some heuristics and trial and error. in the future (if we really wanna dial this in:

- new v1.0.1 rusty kaspa has the sigma wasm code. which has some estimates for storage mass etc.
- handle files better (trim the fat on our handling)
- do some deep diving in to the full encoding process and see where / what we can trim / improve

but for now, this is a large improvement.


HocusLocusTee:
~~@IzioDev note, indexer has been down today (I'm assuming), so I haven't been able to test historical sync flows.~~
Resolved.

IzioDev:
Looks good to me,

To rephrase further possible optimization you mentioned:
We should use a dedicated encoding system that minimizes bloats, JSON isn't suited for that.

Additionally:
We could introduce a compression algorithm

These have already been discussed on Discord, but without real testing, yet :)

PR #222, opened 2025-08-18
Fix-Do/A lot

HocusLocusTee:
## what
- removes utxo-compound modal and makes it part of wallet info (styling etc)
- restyles walletinfo
- refactors a LOT to do with `account-service`, moving a bunch of stuff around, renaming functions, removing password prop passing, changed some of the guards to assert themselves instead. refactor sign and submit into one private helper function, adding fee guard at the start of it etc.
- move the generators to their own service, fixed the generators so the work correctly
- fix ux on fee estimation:
  - if not funds, don't estimate fee (we cant!)
  - when funds come, allow the estimate to kick off again, so in a
  situation where we have text in the box, itll re-hydrate and calc.
- misc styling across the app

### DIFF
--- a/src/service/account-service.ts
+++ b/src/service/account-service.ts
@@ -538,7 +538,6 @@ export class AccountService extends EventEmitter<AccountServiceEvents> {
   }
 
   public async createWithdrawTransaction(
-    // remove the external below and follow payment or compound approacgh
     withdrawTransaction: CreateWithdrawTransactionArgs
   ) {
     this.validateTransactionPrerequisites();
@@ -565,7 +564,7 @@ export class AccountService extends EventEmitter<AccountServiceEvents> {
     }
 
     try {
-      const matureBalance = this.context.balance?.mature ?? 0n;
+      const matureBalance = this.context.balance?.mature;
       const isFullBalance = matureBalance === withdrawTransaction.amount;
       // Use our optimized generator creation method
       const generator = TransactionGeneratorService.createForPaymentOrWithdraw({
@@ -630,8 +629,8 @@ export class AccountService extends EventEmitter<AccountServiceEvents> {
     }
 
     try {
-      const balance = this.context.balance;
-      if (!balance || balance.mature === 0n) {
+      const matureBalance = this.context.balance?.mature;
+      if (!matureBalance || matureBalance === 0n) {
         throw new Error("No mature UTXOs available for compound transaction");
       }
 
@@ -827,7 +826,7 @@ export class AccountService extends EventEmitter<AccountServiceEvents> {
 
     console.log(`Added prefix to address: ${prefixedAddressString}`);
     return new Address(prefixedAddressString);
-  }
+  } // TODO Utility
 
   private isKasiaTransaction(tx: ITransaction | ExplorerTransaction): boolean {
     return tx?.payload?.startsWith(PROTOCOL.prefix.hex) ?? false;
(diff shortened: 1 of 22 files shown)


src/service/transaction-generator.ts
```diff
@@ -0,0 +1,152 @@
...
+    const settings: IGeneratorSettingsObject = {
+      changeAddress: receiveAddress,
+      entries: context,
+      outputs: outputs,
+      payload,
+      networkId,
+      priorityFee: additionalFee,
+    };
+
+    return new Generator(settings);
```
IzioDev, line 81: Should we add a fool guard before returning the generator, to check if `priorityFee` or `priorityFee.amount` (if defined) is greater than an arbitrary value to prevent full drain in case of bad usage?

If accepted, it should also apply to the other 2 static methods 
If accepted, it should invalidate the other arbitrary limit(s) we've set upper layer

src/service/transaction-generator.ts
```diff
@@ -0,0 +1,152 @@
...
+    destinationAddress: Address;
+    amount: bigint;
+    payload: string | Uint8Array;
+    priorityFee?: PriorityFeeConfig;
+  }): Generator {
+    // Check if this is a direct self-message (sending to our own receive address)
+    const isDirectSelfMessage =
+      destinationAddress.toString() === receiveAddress.toString();
+
+    console.log(isDirectSelfMessage);
```
IzioDev, line 36: Is this comment useful? If so, should we instead add log `{isDirectSelfMessage}` to have the context when reading log? (variable name)

Alternatively or additionally, we could log the whole setting object below

src/service/transaction-generator.ts
```diff
@@ -0,0 +1,152 @@
...
+  }: {
+    context: UtxoContext;
+    networkId: string;
+    receiveAddress: Address;
+    destinationAddress: Address;
+    amount: bigint;
+    payload?: string | Uint8Array;
+    priorityFee?: PriorityFeeConfig;
+  }): Generator {
+    const isFullBalance = amount >= (context.balance?.mature ?? 0n);
```
IzioDev: This comment is not applicable to this PR as this is not a new technique, but I think it's something we should research:

Is there a possibility where the balance is not yet resolved to the value? If so, we shouldn't fallback to `0n` and instead throw an error from here probably.

Additionally, we probably shouldn't consider `amount > balance` case (only strict equal), and instead throw an error.

Open question:
* What if the node i'm connected to intentionally lie to me by stating i'm have another balance amount that the reality?

IzioDev (commented):
Looks good overall, better architecture and segregation of concerns 🙏

Open question for UTXO sweep (compound):
* We observe two 'manual' refresh of first UTXO and then of balance. Are they necessary and should we research alternative to get a better controlled / deterministic flow?

src/service/transaction-generator.ts
HocusLocusTee: We already have use of `MAX_TX_FEE` at a lower level than the generator.
Not on every tx type though, ill add it.

src/service/transaction-generator.ts
HocusLocusTee: Nah ill remove. 
I personally think we should be using less logs in the code base, and just log when we need to work on something tricky. There is obviously a case by case basis though.

src/service/transaction-generator.ts
HocusLocusTee: This is something I've been looking into at the moment. Some issues with TN-10, but it has exposed our miss handling of `context.`

HocusLocusTee:
>We observe two 'manual' refresh of first UTXO and then of balance. Are they necessary and should we research alternative to get a better controlled / deterministic flow?


See comment above. Currently looking into this. This is not the correct way to use context.

src/service/transaction-generator.ts
IzioDev: Alright!
We can also think of a proper logger that'd get instrumented on a per-scope basis (automatically, is it possible in JS?) on the long run.

e.g.: `[info] messaging.store - my awesome log here`

src/service/account-service.ts
```diff
@@ -196,6 +212,81 @@ export class AccountService extends EventEmitter<AccountServiceEvents> {
...
+      throw error;
+    }
+  }
+
+  async stop() {
+    console.log("Stopping UTXO subscription and processor...");
+    try {
+      // Stop the UTXO processor
+      await this.processor.stop();
+
```
IzioDev: should we also clear the utxo context? (`this.context.clear()`)

IzioDev (commented):
LGTM!

src/service/account-service.ts
HocusLocusTee: Yep, I've added that.
I also added this to unsub the user so we don't get block events once wallet is closed (as discussed previously)
```js
   // Unsubscribe from block events using the RpcClient method
      if (this.rpcClient.rpc) {
        await this.rpcClient.rpc.unsubscribeBlockAdded();
      }
```

IzioDev:
We can merge this PR whenever you feel comfortable with :)

PR #225, opened 2025-08-22
Handshake v2 - 4. Saved handshakes

IzioDev:
This has become a macro PR:
* refacto app loading strategy, now centralized into a macro hook called `useGlobalStore`, this should simplify the understanding of the client loading flow.
* normalize kasia transaction building mechanism, but now that every caller have the scope context (is one of: message, payment, self stash, handshake), this can be further improved to have a clear entity building per-scope, at a "protocol-app" (just bit above protocol) layer.
* introduce self stash message handling, and more specifically saved handshake. which is a self stash with scope = saved_handshake
* remove block processing logic from account service in favor of a `BlockProcessorService`.
* introduce `SenderAndAcceptanceResolutionService` (saars) that is responsible for tracking acceptance data in-memory, useful to resolve transaction acceptance data and most importantly the sender, without any 3rd party involved!
* both services mentioned are grouped into a zustand `live.store` (it needs to be persistent across app lifecycle)
* add `waitForMatureUtxo` method to the account service, that is useful when wanting to subsequently send multiple transaction
* fixed handshake (both sent & received) ingestion logic that wouldn't update conversation state properly
* simplified historical fetching logic to only handle what's relevant
* fixed indexdb indexes with latest knowledge (anticipate the future lazy loading)
* test: mock wasm modules to allow test involving wasm interraction
* introduce biome configuration (that shouldn't conflict with eslint)

How to test:
* .env:
```
VITE_INDEXER_TESTNET_URL=https://dev-indexer.kasia.fyi/
VITE_DISABLE_INDEXER=false
```
* **Delete indexdb manually**

What's missing:
* handle rpc reconnect (connection loss) and ideally network switch (less minor requirement)


src/hooks/useGlobalStore.ts
```diff
@@ -0,0 +1,116 @@
...
+export class StartSessionInvalidPasswordException extends Error {
+  constructor() {
+    super();
+
+    this.name = StartSessionInvalidPasswordException.name;
+    this.message = "Invalid Password";
+  }
+}
+
+export const useGlobalStore = () => {
```
IzioDev: find a better name

src/components/Modals/LockedSettingsModal.tsx
```diff
@@ -60,45 +61,45 @@ export const LockedSettingsModal: React.FC = () => {
...
       setConnectionSuccess(false);
+    }
+  };
 
-      networkStore.setNetwork(network);
+  const onNetworkChange = (network: NetworkType) => {
+    setConnectionSuccess(false);
 
-      const savedNetwork = localStorage.getItem(`kasia_node_url_${network}`);
+    const savedNetwork = localStorage.getItem(`kasia_node_url_${network}`);
```
IzioDev, line 84: ideally local storage access from react component should be prohibited, and placed at `repository` layer.

src/store/messaging.store.ts
```diff
@@ -162,10 +169,16 @@ export const useMessagingStore = create<MessagingState>((set, g) => {
 
       const knownEventIds = new Set(oooc.events.map((e) => e.id.split("_")[1]));
       const newIndexerPayments = indexerPayments.filter(
-        (p) => !knownEventIds.has(p.tx_id)
+        (p, idx, self) =>
+          !knownEventIds.has(p.tx_id) &&
+          // @todo: remove when indexer dedup filter
```
IzioDev: todo: wait for indexer changes

src/store/messaging.store.ts
```diff
@@ -162,10 +169,16 @@ export const useMessagingStore = create<MessagingState>((set, g) => {
```
IzioDev: todo: payment & message should also use the "after" parameter to optimize historical data loading

IzioDev:
Still a few tiny todo, shouldn't add a lot of code :)

src/store/messaging.store.ts
IzioDev: We can remove these check and consider that https://github.com/K-Kluster/kasia-indexer/pull/16/ will be merged soon

IzioDev:
Of course, it added a lot of code 🤣

So, we've merged the last fixes to prepare a release, with a full network handling refactoring, now considered cleaner.

Waiting for @HocusLocusTee to open the (also big) refacto and broadcast. In the meantime:
1. find a way to host custom wasm properly
2. rollout upgrade indexer + remove client temp patch dedup on the indexer (now not needed)

IzioDev:
I created a temporary Minio (S3-like) for this purpose: https://minio.kasia.fyi/browser/main/

(hosted on our public infrastructure, anonymous read-only access)

PR #227, opened 2025-08-23
Fix/send all and utxo context updates. Walllet Info Style. Tx flow refactor

HocusLocusTee:
## what?

currently the utxocontext + generator don’t gracefully handle sending all funds.

i assume that: with a changeaddress but no actual change, utxocontext never settles and leaves stale `outgoing`. the transfer succeeds, but the context stays wrong.

we tried:

```
context.clear();
context.trackAddresses([this.recv]);
```

after send, but it races balance events and can leave spent funds in `mature`.

what i've settled on is:
after a successful full-balance send, use a new function `retrackAfterFullSend`, which frees the current conteext, then swaps in a fresh utxocontext and re-track the primary address.

seems to work?

also i 
- re added the display types to the wallet (pending, outgoing.
- refactored the processing of generators
- added a warning to withdraw to cold wallet if balance over 30KAS

see handy flow chart of how this all flows:
TxFlow Mermaid Chart-2025-08-23-0101518

### DIFF
--- a/src/components/Modals/WalletInfo.tsx
+++ b/src/components/Modals/WalletInfo.tsx
@@ -33,11 +33,27 @@ export const WalletInfo = () => {
           <h4 className="text-text-primary me-2 font-bold">Wallet Name:</h4>
           <div className="text-base font-semibold">{unlockedWalletName}</div>
         </div>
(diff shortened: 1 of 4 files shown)


IzioDev:
LGTM

PR #230, opened 2025-08-24
Fix/minor style

HocusLocusTee:
## what?

fixes a html error
adjusts some minor coloring and spacing
add display of funds to withdraw modal
renames import success components
restyles migrate wallet component
changes order of utxo context clear and processor stop


IzioDev:
Awesome, thanks!
These console error logs needed to stop 😜

PR #242, opened 2025-09-11
Feat/offchain alias

HocusLocusTee:
## what?

- adds offchain (super duper private) handshakes
- a bit of misc refactoring and style changes
- extracts self stash stuff into their own functions etc

### how?
its a semi hidden super user feature.
long press on the + in directs and youll get a nice little animation on the touch and then the flow will open


src/components/Modals/CopyableValueWithQR.tsx
```diff
@@ -0,0 +1,134 @@
...
+import { Copy, QrCode } from "lucide-react";
+import { Button } from "../Common/Button";
+import { toast } from "../../utils/toast-helper";
+import { copyToClipboard } from "../../utils/clipboard";
+
+type CopyableValueWithQRProps = {
+  value?: string;
+  label?: string;
+  qrTitle?: string;
+  sectionId?: string; // NOTE lets not pass this in, lets just do it in the parent.
```
IzioDev: I don't understand the comment, should we enforce this props?

src/components/Modals/NewChatForm.tsx
```diff
@@ -231,19 +245,15 @@ export const NewChatForm: React.FC<NewChatFormProps> = ({ onClose }) => {
...
-      } else {
-        text = await navigator.clipboard.readText();
-      }
-
-      setRecipientInputValue(text.toLowerCase());
-    } catch {
-      // Handle clipboard access error silently or show a toast
-      console.warn("Failed to paste from clipboard");
+    let text = "";
+    if ("__TAURI_INTERNALS__" in window) {
```
IzioDev: renamed `core.isTauri` imported from tauri package (just as a reminder, because it will generate conflicts

src/components/Modals/OffChainHandshakeModal.tsx
```diff
@@ -0,0 +1,616 @@
...
+    createSelfStash,
+    setOpenedRecipient,
+  } = useMessagingStore();
+  const { openModal, setQrScannerCallback, modals } = useUiStore();
+  const balanceMature = useWalletStore((s) => s.balance?.mature);
+
+  // check if user has sufficient funds for self stash (0.2 KAS minimum)
+  const hasSufficientFundsForSelfStash = useMemo(() => {
+    return balanceMature ? balanceMature >= BigInt(20000000) : false;
+  }, [balanceMature]);
```
IzioDev: self stash should not require 0.2 Kas, but at least network fees

src/service/conversation-manager-service.ts
```diff
@@ -508,9 +509,9 @@ export class ConversationManagerService {
           const _conversation: Conversation = {
             id: uuidv4(),
             myAlias: payload.alias,
-            theirAlias: null,
+            theirAlias: payload.theirAlias || null, // use theirAlias if present (for offline handshakes)
```
IzioDev, line 512: nice catch:
their alias could also be present in the following case:
* i received handshake that I accept
* i create saved handshake with "theirAlias" populated
* later, i connect on a new device, saved handshake gets populated with their alias

it's optional in previous cases because it would have been "overwritten" by received handshake which isn't the case here

src/store/messaging.store.ts
```diff
@@ -101,13 +101,30 @@ interface MessagingState {
...
   // Last opened recipient management
   restoreLastOpenedRecipient: (walletAddress: string) => void;
 
+  // self stash management
+  createSelfStash: (handshakeData: {
+    type: "initiation" | "response";
+    partnerAddress: string;
+    ourAlias: string;
+    theirAlias?: string;
+    recipientAddress?: string;
```
IzioDev: What is the difference between partnerAddress and recipientAddress?

src/store/messaging.store.ts
```diff
@@ -1715,5 +1691,92 @@ export const useMessagingStore = create<MessagingState>((set, g) => {
...
+          theirAlias: handshakeData.theirAlias,
+        }),
+        ...(handshakeData.isResponse && { isResponse: true }),
+      };
+
+      // create extended payload for self-stash storage
+      const payload = {
+        ...basePayload,
+        partnerAddress: handshakeData.partnerAddress,
+        recipientAddress: handshakeData.recipientAddress || address,
```
IzioDev: I think we should either guess from currently loaded address or get the source of truth from parameter but I find that mixing is confusing, up to you to keep it as is or not

IzioDev (commented):
Awesome, love the idea behind the feature

src/components/Modals/CopyableValueWithQR.tsx
HocusLocusTee: That was a note to myself. I should've left a TODO. This was from the initial version of the component.

HocusLocusTee: This has been fixed now.

src/components/Modals/OffChainHandshakeModal.tsx
HocusLocusTee: Changing this and leaving it as 0.19 kas for now. Eventually I want to make some functions to do balance checks (returning true or false and not passing the balance around).
I will fine tune this then.

src/store/messaging.store.ts
HocusLocusTee: Nothing pragmatically. Its an artifact. Cleaned up.

src/store/messaging.store.ts
HocusLocusTee: Fixed.

HocusLocusTee:
@IzioDev ready to review again or merge.

PR #244, opened 2025-09-17
Feat/password use

HocusLocusTee:
## what?

adds a security feature that checks when password was last entered on value tx over a set amount (5 kas currently).
if over a time frame (10 min currently) we ask for reauth.
it wont stop users from sending messages, but payments and withdrawals only.

can also be bypassed in feature flips.

minor: makes a utility function to get TKAS or KAS based on network type. wasm does also have this but ...

## to test:
go to constants and set new constant to like 0.1 min. then try to make a payment.


HocusLocusTee:
closing this as its uninteresting for now

PR #246, opened 2025-09-23
fix: routing to wallet diretly causes no wallet selection

HocusLocusTee:
## what?

routing to wallet id directly currently does not select wallet id to unlock.
this fixes it.

_note: maybe it was an issue with the new branch only. but I didn't confirm - this fixes it regardless._


IzioDev:
I confirm this was an existing bug on the actual version, nice one!

PR #247, opened 2025-09-23
fix(historical): fetching checkpoints

IzioDev:


src/store/repository/db.ts
```diff
@@ -325,6 +327,8 @@ export class Repositories {
       tenantId,
       walletPassword
     );
+
+    this.metadataRepository = new MetaRespository(tenantId);
```
HocusLocusTee, line 331: Since this uses `localstorage` and not indexeddb, should we just init this in `useOrchastrator` instead?

Or do you want to make this a _general data access_. The name implies this  - `Repositories` - but its current use is db only.

HocusLocusTee (approved):
Works as intended.
Comment below is not a blocker.

src/store/repository/db.ts
IzioDev: That's a fair point to raise,

For now, I just wanted a way to have centralized data accessor initiators, to have a better tracking.

I'm unsure about the perfect end structure, but I guess we could have a separate "Repositories" class like "KVStore" or something of this sort. What do you think?

IzioDev:
I'm merging this, but it doesn't mean i skip the comment, let's keep discussing it on Discord

PR #249, opened 2025-09-23
Feat/show password icon

IzioDev:
image

### DIFF
--- a/src/store/session.store.ts
+++ b/src/store/session.store.ts
@@ -72,7 +72,7 @@ export const useSessionState = create<SessionState>((set, get) => {
     },
     async hasSession(tenantId) {
       // temporary disable for iOS
-      if (platform() === "ios") {
+      if (!core.isTauri() || platform() === "ios") {
         return false;
       }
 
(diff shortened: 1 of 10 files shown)


src/store/session.store.ts
```diff
@@ -87,7 +87,7 @@ export const useSessionState = create<SessionState>((set, get) => {
     },
     async setSession(tenantId, password) {
       // temporary disable for iOS
-      if (platform() === "ios") {
+      if (!core.isTauri() || platform() === "ios") {
```
IzioDev, line 90: Undetected bug on web version, my bad, this fixes it.

HocusLocusTee: Ahh I just saw this.

HocusLocusTee:
image

Some browsers (like edge), do this already.

Global solution is to add:
`input[type="password"]::-webkit-reveal-button {
  display: none;
}`
to our `index.css`.

Or you can reference it in tailwin on the PasswordField component.

HocusLocusTee (approved):
Other than my comment above... looks and feels good to me!

IzioDev:
> image
> 
> Some browsers (like edge), do this already.
> 
> Global solution is to add: `input[type="password"]::-webkit-reveal-button { display: none; }` to our `index.css`.
> 
> Or you can reference it in tailwin on the PasswordField component.

Done, nice catch

PR #250, opened 2025-09-23
feat(native): support camera access

IzioDev:
image
image


HocusLocusTee:
Best to just merge this and test with beta testers imo.

PR #251, opened 2025-09-23
fix: remove wallet legacy mention completely

IzioDev:
image
image
image


HocusLocusTee:
This branch re-labels all of the existing 'legacy' wallets to standard but derives a different address.
If there a user out there that has funds on this, they will be 'lost' (within kasia). 
 
Not a big deal imo for following reasons:
- they still have wallet access and can export seed to derive somewhere else (we can assist if needed)
- an assumption that no one is A: currently using kasia, B: using kasia 'legacy wallets' to store funds

As long as this is intended by you (I'm onboard), then LGTM

HocusLocusTee (approved):

IzioDev:
> This branch re-labels all of the existing 'legacy' wallets to standard but derives a different address. If there a user out there that has funds on this, they will be 'lost' (within kasia).
> 
> Not a big deal imo for following reasons:
> 
>     * they still have wallet access and can export seed to derive somewhere else (we can assist if needed)
> 
>     * an assumption that no one is A: currently using kasia, B: using kasia 'legacy wallets' to store funds
> 
> 
> As long as this is intended by you (I'm onboard), then LGTM

Your thoughts were also mine at that time.

It was intentional indeed, we could always add a tooltip near wallet balance display, what do you think?

HocusLocusTee:
>It was intentional indeed, we could always add a tooltip near wallet balance display, what do you think?

Imo not a big deal. For the potential 0-1 users that has issues, I think we just support them via discord.

PR #256, opened 2025-09-26
refactor(style) + fix some things

HocusLocusTee:
## what?
refactor of global style and fonts

- alter mode selection (bcast and Dms): if only 2 modes enabled (current max)  that just instant switch with a toast
- fix misc styling bugs
- introduce a 'delete' wallet modal. reason: mobiles didn't respect the window alter and delete does not show
- alter delete all messages and give a sub pane in settings modal
- alter payment popup on mobile to be full width and correct size
- adjust broadcasts so they init on wallet load (lazy) so we can listen for them as long as the app is open
- adjusts the way we prevent mobile zoom

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 .rules
-
+.cargo
 wasm/*
 
 # Logs
--- a/index.html
+++ b/index.html
@@ -4,7 +4,7 @@
     <meta charset="UTF-8" />
     <meta
       name="viewport"
-      content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no, viewport-fit=cover"
+      content="width=device-width, initial-scale=1.0, viewport-fit=cover"
     />
     <link rel="icon" type="image/png" href="/kasia-logo.png" />
     <link rel="stylesheet" href="/src/index.css" />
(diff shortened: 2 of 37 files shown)


HocusLocusTee:
`f2997891700de4ee6118df5ecff6d090a0afa6f5` - adjusts the way we prevent mobile zoom.
Need to make sure tauri stuff still operates as expected.
@IzioDev

IzioDev:
> `f2997891700de4ee6118df5ecff6d090a0afa6f5` - adjusts the way we prevent mobile zoom. Need to make sure tauri stuff still operates as expected. @IzioDev

Is this related to zoom happening while double tapping while having the devtool opens?
If so, this is a very old bug that is being carried since 10 years+. If the workaround works, this would be huge.

HocusLocusTee:
> > `f2997891700de4ee6118df5ecff6d090a0afa6f5` - adjusts the way we prevent mobile zoom. Need to make sure tauri stuff still operates as expected. @IzioDev
> 
> 
> 
> Is this related to zoom happening while double tapping while having the devtool opens?
> 
> If so, this is a very old bug that is being carried since 10 years+. If the workaround works, this would be huge.

Nay not quite.
It disables pinch and double tap zoom (at least on iOS-web).
Downside is that IF the browser auto zooms - due to a small px on an input - you have to reload the app to clear.
But, I think I've resized most of the problematic inputs.

PR #260, opened 2025-10-04
Fix/misc

HocusLocusTee:
## what?
- tries to fix pwas scrolling
- fixes one piece of copy that was new lining on mobile
- style: moves the contact menu to the end on mobile, see below:
- add chevron to left of the address
- change copy from "Create New Wallet" > "Create Wallet"
- fixes fee estimation not working or showing on Image attachments (in Directs)
- alter priority fee to not close on selection, but close on button
- if you click the top left logo on desktop, you will route home
- reduces fee estimate to every few seconds
- adds wait time to fee display IF >0.5s
- adds donation component in locked settings
- increases touch area on send buttons
- bump version to 0.6.2

### DIFF
--- a/src/containers/MessengerProvider.tsx
+++ b/src/containers/MessengerProvider.tsx
@@ -220,47 +220,45 @@ export const MessengerProvider: FC = () => {
   return (
     <>
       {/* Main Message Section*/}
-      <div className="bg-primary-bg flex items-center">
-        <div className="flex h-[100dvh] w-full overflow-hidden sm:h-[calc(100dvh-69px)]">
-          {isWalletReady &&
-          messageStore.isLoaded &&
-          walletStore.isAccountServiceRunning ? (
-            <>
-              <SidebarSection
-                onContactClicked={onContactClicked}
-                onModeChange={onModeChange}
-                openedRecipient={messageStore.openedRecipient}
-                walletAddress={walletStore.address?.toString()}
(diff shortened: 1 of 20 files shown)


src/containers/MessengerProvider.tsx
```diff
@@ -220,47 +220,45 @@ export const MessengerProvider: FC = () => {
...
-                setMobileView={(view) => {
-                  setMobileView(view);
-                  // when switching to contacts view on mobile, navigate back to the appropriate base route
-                  if (isMobile && view === "contacts") {
-                    if (isCurrentlyInBroadcastMode) {
-                      navigate(`/${walletId}/broadcasts`);
-                    } else {
-                      navigate(`/${walletId}/directs`);
-                    }
+      <div className="bg-primary-bg flex h-[100dvh] overflow-hidden sm:h-[calc(100vh-69px)] pointer-coarse:h-[calc(100dvh-var(--kb,0px))]">
```
HocusLocusTee, line 223: Everything else in the diff is noise.
The fix is this here:
`pointer-coarse:h-[calc(100dvh-var(--kb,0px))]`

Its gritty but since ios26 has this new floating glass thing above the keyboard - dvh accounts for this and ends up breaking everything.

I've used pointer-coarse as SOME way to see that we are on mobile (not just small breakpoints).

Needs testing on tauri and android pwa (sorry)

PR #265, opened 2025-10-10
chore(web): vite pwa cache policy tweak

IzioDev:
This should fix stale cache

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "kasia",
-  "version": "0.6.1",
+  "version": "0.6.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "kasia",
-      "version": "0.6.1",
+      "version": "0.6.2",
       "dependencies": {
         "@headlessui/react": "2.2.1",
         "@tailwindcss/vite": "4.1.3",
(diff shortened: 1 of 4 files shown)


HocusLocusTee:
I had caching issues on browser (not just PWA).
We need probably need to make sure the index.html is not getting cached (and the generated service worker?)

We should be able to set this quite easily with a `_headers` file as specified [here](https://developers.cloudflare.com/pages/configuration/headers/)

IzioDev:
> I had caching issues on browser (not just PWA). We need probably need to make sure the index.html is not getting cached (and the generated service worker?)
> 
> We should be able to set this quite easily with a `_headers` file as specified [here](https://developers.cloudflare.com/pages/configuration/headers/)

Updated with this method

HocusLocusTee:
Nice.

We could add this too?
```
/*.wasm
  Cache-Control: public, max-age=31536000, immutable

/*.png
  Cache-Control: public, max-age=604800

/*.svg
  Cache-Control: public, max-age=604800
  ```

IzioDev:
> Nice.
> 
> We could add this too?
> 
> ```
> /*.wasm
>   Cache-Control: public, max-age=31536000, immutable
> 
> /*.png
>   Cache-Control: public, max-age=604800
> 
> /*.svg
>   Cache-Control: public, max-age=604800
> ```

Yes for png and svg 👍 
Porbably yes also for wasm if we use a specific wasm file name, in case we update SDKs (as we did recently). One way of "fixing" this would be to use `kaspa-${version}.wasm`

HocusLocusTee:
> > Nice.
> > We could add this too?
> > ```
> > /*.wasm
> >   Cache-Control: public, max-age=31536000, immutable
> > 
> > /*.png
> >   Cache-Control: public, max-age=604800
> > 
> > /*.svg
> >   Cache-Control: public, max-age=604800
> > ```
> 
> Yes for png and svg 👍 Porbably yes also for wasm if we use a specific wasm file name, in case we update SDKs (as we did recently). One way of "fixing" this would be to use `kaspa-${version}.wasm`

True. Up to you, I don't mind excluding it for now.

PR #266, opened 2025-10-11
style: new toast styling + more

HocusLocusTee:
## what?
re-designs the presentation of the toasts (from being just a screen list) to having some depth.
shows a max of 3 at a time. gives some nice(r) animations etc.

image

## test:
add something like this to message provider to spam a whole bunch of toasts:

```
  useEffect(() => {
    const toastTypes = ["success", "error", "warning", "info"];

    const generateRandomMessage = () => {
      const length = Math.floor(Math.random() * (100 - 5 + 1)) + 5;
      const chars =
        "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789 ";
      let result = "";
      for (let i = 0; i < length; i++) {
        result += chars.charAt(Math.floor(Math.random() * chars.length));
      }
      return result;
    };

    const interval = setInterval(() => {
      const randomType =
        toastTypes[Math.floor(Math.random() * toastTypes.length)];
      toast[randomType](generateRandomMessage());
    }, 1000);
    return () => clearInterval(interval);
  }, []);
```

## what else?
adds donation address to the logged in settings screen
adds toast for message import import (initiate and completion)
adjust hold-able + button
remove toast removal on all route changes
reduce toast duration on mode switch

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -68,6 +68,20 @@
   --color-kas-primary: var(--kas-primary);
   --color-kas-secondary: var(--kas-secondary);
   --color-text-warning: var(--text-warning);
+
+  --animate-slide: slideInFromRight 0.5s ease-out;
+
+  @keyframes slideInFromRight {
+    from {
+      opacity: 0;
+      transform: translateX(40%);
+    }
+
+    to {
+      opacity: 1;
+      transform: translateX(0);
+    }
+  }
 }
 
 /* we need this to assist with splash > loaded app transition. Its our css limit */
@@ -112,18 +126,6 @@ input[type="password"]::-webkit-reveal-button {
 
   to {
     opacity: 1;
-  }
-}
-
-@keyframes slideInFromRight {
-  from {
-    opacity: 0;
-    transform: translateX(40%);
-  }
-
-  to {
-    opacity: 1;
-    transform: translateX(0);
   }
 }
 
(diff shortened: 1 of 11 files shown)


src/index.css
```diff
@@ -115,6 +115,18 @@ input[type="password"]::-webkit-reveal-button {
...
+  from {
+    opacity: 0;
+    transform: translateX(40%);
+  }
+
+  to {
+    opacity: 1;
+    transform: translateX(0);
+  }
+}
```
IzioDev: Minor, but apparently there's a more integrated way to integrate custom TW animations: https://tailwindcss.com/docs/animation#customizing-your-theme

IzioDev:
Feel free to ignore and merge if you find this method not interesting

src/index.css
HocusLocusTee: Nice, this seems better.
Ill make this change then explore one more bug (+ not working without a hold on some devices) then merge.

HocusLocusTee:
@IzioDev Made a few more changes.
```
adjust hold-able + button
remove toast removal on all route changes
reduce toast duration on mode switch
```

note the middle change `remove toast removal on all route changes` now that we have new toasting we dont need this (also this was put in before the new routing changes).

once you ack this, feel free to merge.

PR #267, opened 2025-10-12
chore: optimize dev loading time and support indexer-less app

IzioDev:


scripts/run-vite-with-env.cjs
```diff
@@ -39,7 +39,7 @@ let fullCommand = "";
 
 switch (command) {
   case "dev":
-    fullCommand = `vite ${viteCommand}`;
+    fullCommand = `vite --debug ${viteCommand}`;
```
HocusLocusTee: ✨

package.json
```diff
@@ -56,7 +56,7 @@
     "@types/react": "19.1.0",
     "@types/react-dom": "19.1.4",
     "@types/uuid": "^10.0.0",
-    "@vitejs/plugin-react": "^4.5.1",
+    "@vitejs/plugin-react-swc": "^4.1.0",
```
HocusLocusTee, line 59: builds v-qik now

HocusLocusTee:
Initial feedback based on some testing - this being *general* as I haven't paid to attention to this for a while:
 - post unlock there is a lot of UI thread-blocking if indexer hydration is large
 - once this happened: messages from indexer 'load' into the chat very slowly, I was getting one per second (even though indexer returns 100)
 
 - odd behavior: after deleting db and unlocking a few times the indexer stopped returning my indexed messages and no only returns handshakes

HocusLocusTee:
Unrelated to your changes (but related to stuff being slow). We should change this from trace to log:
https://github.com/K-Kluster/Kasia/blob/78934b8415e643bb8f30742734b89702ee2a2b5c/src/store/messaging.store.ts#L474

trace is generally a fair bit slower.

IzioDev:
> Initial feedback based on some testing - this being _general_ as I haven't paid to attention to this for a while:
> 
>     * post unlock there is a lot of UI thread-blocking if indexer hydration is large
> 
>     * once this happened: messages from indexer 'load' into the chat very slowly, I was getting one per second (even though indexer returns 100)
> 
>     * odd behavior: after deleting db and unlocking a few times the indexer stopped returning my indexed messages and no only returns handshakes

1. post unlock, only handshake hydration should be blocking.
2. did you have the dev tools opened? I found it to be a potential cause of slowness. And in theory, it should only persist (in store) once everything is resolved, if not, it's a bug

1&2. indexer hydration should only operate on non known events, if somehow it also include already known event, it's a bug

3. this is confirmed to be a bug, delete db (from the ui) does not delete the local storage metadata corresponding to this wallet (checkpoints)

HocusLocusTee:
Yeh, I noticed 1 and 2 on separate occasions. But they likely were the same issue.

No re-hydrating on existing events.

RE:Point 3, you might have been doing staging index updating at the same time because not long after - you posted saying you had updated the infra.

PR #269, opened 2025-10-17
Feat/blocklist

HocusLocusTee:
## what?
initial 'block a contact' feature

adds:
ability to block from broadcasts or directs contact info pane
ability to see blocked addresses and a broadcast vis option in settings

stores the blocked address in db (also has the capacity for a 'reason') - note: my initial version included a reason field and allowed users to block an address straight from the settings, but I removed this as it was a little complicated handling the visibility states etc. So atm, we have no UI to add a reason and no way to block a contact unless its on their contact card.

blocking will delete any in db contact or messages. BUT in bcasts you will either still see a "Blocked Contact" message or no message (depending on your settings)

also adds `kasia-general` as a default initial channel. users can block now so this seems reasonable?


src/store/messaging.store.ts
IzioDev, line 1: inside `flushWalletHistory`, should we allow also delete blocking list?

if not, `deleteTenant` on the `blockedAddress` repository seems not useful

src/components/Modals/ContactInfoModal.tsx
```diff
@@ -2,84 +2,148 @@ import { FC } from "react";
...
-        </div>
-      </div>
-      {/* Indented content below avatar/nickname/contact */}
-      <div className="space-y-2 pl-2">
-        {" "}
-        {/* pl-14 aligns with avatar+gap */}
-        <div>
-          <div className="text-xs font-medium tracking-wide text-[var(--text-secondary)] uppercase">
-            Address
+export const ContactInfoModal: FC<ContactInfoModalProps> = ({
```
IzioDev, line 12: What do you think on placing the block button here instead?

Image

src/store/messaging.store.ts
```diff
@@ -527,6 +528,15 @@ export const useMessagingStore = create<MessagingState>((set, g) => {
...
 
       const processOneSender = async (senderAddress: string) => {
+        // Skip processing historical data for blocked addresses
+        const blocklistStore = useBlocklistStore.getState();
+        if (blocklistStore.blockedAddresses.has(senderAddress)) {
+          console.log(
+            `Skipping historical handshake processing for blocked address: ${senderAddress}`
+          );
+          return;
+        }
```
IzioDev: I think we can filter out a bit earlier in the flow to avoid potentially expensive computation (decryption).

HistoricalSyncer could take blockedAddress repository as parameter and filter out ignored participants (as early as possible). Or, if you want messaging store to still keep this responsibility, one filter per direction (sent/received) in each loop (line 413 and line 462)

IzioDev (changes requested):

src/components/Modals/ContactInfoModal.tsx
HocusLocusTee: Blocking people doesn't seem too frequent, unless in broadcast and that has its own modal anyway.
What are you concerns?

IzioDev: My only concern would be people not finding the option (dropdown + modal + dropdown)

I see people still having trouble to discover you have the ability to rename your contacts. But I'm fine with proposed changes, what just gathering your thoughts

HocusLocusTee: I have pushed changes with this. it does feel better. The modal warns people about whats going to happen, so I feel ok with the primary position.

Deleting wallet history deletes a blocked info too.

Also added import export ability.

src/store/messaging.store.ts
HocusLocusTee: With the rebase, this has been solved.

IzioDev:
There are conflicts with base branch. Once resolved, I'll checkout locally to push it asap (before we start a protocol change)

HocusLocusTee:
> There are conflicts with base branch. Once resolved, I'll checkout locally to push it asap (before we start a protocol change)

Conflicts have been dealt with.

IzioDev:
One unexpected behavior:
1. have an existing conversation from a to b
2. a blocks b
3. a unblocks b
4. a don't see b's conversation (can be ok ux-wise), but when initiating an handshake, app says: "conversation already exists with [...]"

HocusLocusTee:
> One unexpected behavior:
> 
> 1. have an existing conversation from a to b
> 
> 2. a blocks b
> 
> 3. a unblocks b
> 
> 4. a don't see b's conversation (can be ok ux-wise), but when initiating an handshake, app says: "conversation already exists with [...]"

Is it an issue if you log out and then back in?

IzioDev:
> > One unexpected behavior:
> > 
> > 1. have an existing conversation from a to b
> > 2. a blocks b
> > 3. a unblocks b
> > 4. a don't see b's conversation (can be ok ux-wise), but when initiating an handshake, app says: "conversation already exists with [...]"
> 
> Is it an issue if you log out and then back in?

On logout/login back (without browser refresh), I can initiate the conversation (internal state has been flushed in the meantime).

HocusLocusTee:
> > > One unexpected behavior:
> > > 
> > > 1. have an existing conversation from a to b
> > > 2. a blocks b
> > > 3. a unblocks b
> > > 4. a don't see b's conversation (can be ok ux-wise), but when initiating an handshake, app says: "conversation already exists with [...]"
> > 
> > 
> > Is it an issue if you log out and then back in?
> 
> On logout/login back (without browser refresh), I can initiate the conversation (internal state has been flushed in the meantime).

This has been fixed. Likely itll be useless and confusing until can have deterministic alias (as re adding on your end give you another alias). But since our next release should include alias changes, its probably fine as is.

IzioDev:
Tested locally, mergeable (fix conflicts needed first).
Only flaw, but not related to blocklist directly:
- After blocking and unblocking someone, if I send handshake and a message it works (b sees my message), but when b send message, I don't see it

So I'd say we merge and re-test after deterministic alias

PR #270, opened 2025-10-19
Style/add indexer settings

HocusLocusTee:
## what?

reworks locked settings to allow manual indexer url
disable indexer all together

moves these connection settings to a utility + local storage, removes them from the env file

when indexer is disabled, adds some info in the header or slide menu to show that

### DIFF
--- a/.env.production
+++ b/.env.production
@@ -8,7 +8,6 @@ VITE_DEV_MODE=false
 
 VITE_INDEXER_MAINNET_URL=https://indexer.kasia.fyi/
 VITE_INDEXER_TESTNET_URL=https://dev-indexer.kasia.fyi/
-VITE_DISABLE_INDEXER=false
 
 VITE_DEFAULT_MAINNET_KASPA_NODE_URL=wss://wrpc.kasia.fyi
 VITE_DEFAULT_TESTNET_KASPA_NODE_URL=wss://dev-wrpc.kasia.fyi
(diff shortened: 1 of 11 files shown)


HocusLocusTee:
Note @izio, while testing this I got to test your underlying PR changes. No issues encountered.

IzioDev:
Small detail, the container isn't centered vertically on desktop (it is on mobile)
image

IzioDev:
minor:
* open the settings page > node settings
* manual > enter a wrong address and save, we're now disconnected
* switch to auto
=> we're still disconnected

HocusLocusTee:
> minor:
> 
>     * open the settings page > node settings
> 
>     * manual > enter a wrong address and save, we're now disconnected
> 
>     * switch to auto
>       => we're still disconnected

There was a more fundamental issue which I think is now fixed.

Also added the ability to test indexer from the locked settings.
Lmk if you want to use part of the generated api instead of fetches.

IzioDev:
> > minor:
> > ```
> > * open the settings page > node settings
> > 
> > * manual > enter a wrong address and save, we're now disconnected
> > 
> > * switch to auto
> >   => we're still disconnected
> > ```
> 
> There was a more fundamental issue which I think is now fixed.
> 
> Also added the ability to test indexer from the locked settings. Lmk if you want to use part of the generated api instead of fetches.

If using the generated API is too messy in the context of setting the URL up (I think it has to be globally overwritten by our current architecture), i'm fine with fetches

lgtm

PR #271, opened 2025-10-20
Tmp/editable aliases

HocusLocusTee:
## what?
adds the ability for a user to edit their aliases from the contact info modal
extracts self stash to a hook

fixes a bug (that was reintroduced) where the client processes messages when logged in then logged out again

some thoughts while working on this:
nearly everyone could use the same alias and just rely on the client to successfully or not successfully decrypt, encrypt the message - the impact of this requires some fundamental protocol thought from me. 

additionally this also gives the ability for on-chain handshake obfuscation. where party A and party B could start a regular chat sending a transaction handshake to each other. 
then within the encrypted chat they could agree on new aliases and swap those. a third party could assume that this has taken place but its not provable.

### DIFF
--- a/src/service/block-processor-service.ts
+++ b/src/service/block-processor-service.ts
@@ -36,9 +36,10 @@ export class BlockProcessorService extends EventEmitter<{
   // ordering matters, Set is safe as per MDN documentation
   // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set
   private processedTransactionIds: Set<string> = new Set();
-  private monitoredConversations: Set<string> = new Set(); // Store monitored aliases
-  private monitoredAddresses: Map<string, string> = new Map(); // Store address -> alias mappings
-  private readonly MAX_TRANSACTION_IDS_RETENTION_COUNT = 1_000; // Prevent unlimited growth
+  private monitoredConversations: Set<string> = new Set();
+  private monitoredAddresses: Map<string, string> = new Map();
+  private readonly MAX_TRANSACTION_IDS_RETENTION_COUNT = 1_000;
+  private boundProcessBlockAdded: (event: IBlockAdded) => Promise<void>;
 
   constructor(
     private readonly rpc: RpcClient,
@@ -47,19 +48,18 @@ export class BlockProcessorService extends EventEmitter<{
   ) {
     super();
 
+    this.boundProcessBlockAdded = this.processBlockAdded.bind(this);
     this.init();
   }
 
   private async init() {
-    this.rpc.addEventListener("block-added", this.processBlockAdded.bind(this));
+    this.rpc.addEventListener("block-added", this.boundProcessBlockAdded);
     await this.rpc.subscribeBlockAdded();
   }
 
-  stop() {
-    this.rpc.removeEventListener(
-      "block-added",
-      this.processBlockAdded.bind(this)
-    );
+  async stop() {
+    this.rpc.removeEventListener("block-added", this.boundProcessBlockAdded);
+    await this.rpc.unsubscribeBlockAdded();
   }
 
   private async processBlockAdded(_event: IBlockAdded) {
--- a/src/store/messaging.store.ts
+++ b/src/store/messaging.store.ts
@@ -49,6 +49,7 @@ import {
 } from "../service/import-export-service";
 import { useNetworkStore } from "./network.store";
 import { historicalLoader_loadSendAndReceivedHandshake } from "../utils/historical-loader";
+import { validateAlias } from "../utils/alias-validator";
 
 interface MessagingState {
   isLoaded: boolean;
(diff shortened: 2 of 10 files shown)


IzioDev:
>  nearly everyone could use the same alias and just rely on the client to successfully or not successfully decrypt, encrypt the message - the impact of this requires some fundamental protocol thought from me.

Kasia in its early days was kind of working like this: process every messages and try to decrypt, whether or not the alias was known. This added extra slowness randomly while using the application (background listening + decrypt)

src/service/block-processor-service.ts
```diff
@@ -47,19 +48,18 @@ export class BlockProcessorService extends EventEmitter<{
...
     await this.rpc.subscribeBlockAdded();
   }
 
-  stop() {
-    this.rpc.removeEventListener(
-      "block-added",
-      this.processBlockAdded.bind(this)
-    );
+  async stop() {
+    this.rpc.removeEventListener("block-added", this.boundProcessBlockAdded);
```
IzioDev, line 61: Are we implying this line isn't working as intended?

src/store/messaging.store.ts
```diff
@@ -595,6 +603,11 @@ export const useMessagingStore = create<MessagingState>((set, g) => {
             contact: conversationWithContact.contact,
             events: [],
           };
+        } else {
+          // update the conversation reference with the latest from manager
+          // this ensures alias updates from self-stash are reflected (newer than db aliases)
+          oooc.conversation = conversationWithContact.conversation;
+          oooc.contact = conversationWithContact.contact;
```
IzioDev: todo izio: validate this flow and see if we shouldn't handle it elsewhere

note: it is expected (not necessarily working in every case) that this whole logic takes the last knows alias based on date of events, if there are multiple receive / sent for a single recipient, we should take the last known one, regardless

HocusLocusTee: Note: without this we would save the new alias. Stash it. Log out and back in but contact info model would reference old alias. So
You could either edit it which would update the state or just log out and back in again.
Initial self stash is handles already in logic above this line.

src/service/block-processor-service.ts
HocusLocusTee: Are you talking about this line or
in general? Observed behaviour before this is we still process blocks when logged out (after logging in). Unsubbing for block added is the main fix

IzioDev: Yes this line,
So based on your comment it seems like `this.rpc.removeEventListener("block-added", this.boundProcessBlockAdded);` isn't working, but still your added line makes sense.

It's just that the previous one should stop incoming block for being processed, while the other one should tell block processor to stop ingest them (but they still are incoming)

I'll have a look 👍

src/store/messaging.store.ts
IzioDev: > Stash it. Log out and back in but contact info model would reference old alias

This likely indicate a greater issue where newer handshake (received / sent) events on this conversation aren't being ingested properly, potentially the same happens if I receive a new handshake while already having a conversation, we'd like the new handshake to be the new source of truth

src/service/block-processor-service.ts
IzioDev: Could you please annotate the line (`await this.rpc.unsubscribeBlockAdded();`) with a TODO, like: This shouldn't be necessary as the listener on `block-added` is supposed to be removed (which seems to not be the case), so we force the un subscription from internal system

IzioDev:
After this comment https://github.com/K-Kluster/Kasia/pull/271#discussion_r2612916826, mergeable 🚀

HocusLocusTee:
> After this comment [#271 (comment)](https://github.com/K-Kluster/Kasia/pull/271#discussion_r2612916826), mergeable 🚀

I still not quite understand. Irrespective of the 'bug' of us receiving blocks after removing event listener - why would we not want to unsub? 

Note: I had to re-base on staging.

IzioDev:
> > After this comment [#271 (comment)](https://github.com/K-Kluster/Kasia/pull/271#discussion_r2612916826), mergeable 🚀
> 
> I still not quite understand. Irrespective of the 'bug' of us receiving blocks after removing event listener - why would we not want to unsub?
> 
> Note: I had to re-base on staging.

We do want to unsub, it's the correct line to add. But it's also has reflected an underlying bug

IzioDev:
Modified so we allow editing their alias even if N/A (unset)

If you're satisfied with the change, LGTM @HocusLocusTee

HocusLocusTee:
> Modified so we allow editing their alias even if N/A (unset)
> 
> 
> 
> If you're satisfied with the change, LGTM @HocusLocusTee 

I was planning to not merge this. And just pick some relevant parts after deterministic alias is added.

IzioDev:
All right, good for me! Should we mark it draft?

Issue #272, opened 2025-11-13
Improvements for Consideration


MEZTech-LLC:
Low hanging fruit: 
- Improve the image sending functionality. We should allow users to configure the compression to some extend (a simple scalar/slidebar and feedback on the filesize and/or the fee is probably sufficient). 
-Improve image viewer. On desktop browser, images received seem pretty small (not sure if this is a result of resizing during compression - if that's the case, we should definitely limit the degree of scaling done during compression and rely more on JPEG cutoff frequencies or Webp specific params) , we should expand these. 
-Download decoded images button.

Slightly more difficult: 
- Video encoding/decoding
- Audio only encoding/decoding

Wondering if there's any ongoing work or interest in implementing any of the above.

MEZTech-LLC:
> [KAS-66](https://linear.app/klusterkasia/issue/KAS-66/improvements-for-consideration)

No access

HocusLocusTee:
> > [KAS-66](https://linear.app/klusterkasia/issue/KAS-66/improvements-for-consideration)
> 
> No access

You would need to be added. Linear doesn't run publicly. Comments and discussions are on GitHub, linear is more so for (my) personal tracking. I can give you access if you'd like to start contributing.

MEZTech-LLC:
- Delete chats should be a feature (important to allow retries on failed handshakes -- currently only seems possible by deleting all messages in the account).

HocusLocusTee:
Delete chats is basically done here. https://github.com/K-Kluster/Kasia/pull/269.
There is no real distinction between blocking a user and deleting one chat.

MEZTech-LLC:
I'd argue that from a user's perspective there's a difference. 
From a user's point of view, delete chat should remove chat history (and this chat history should ideally be deleted from the db and app cache); but handshake shouldn't have to be remade.

HocusLocusTee:
> I'd argue that from a user's perspective there's a difference. 
> From a user's point of view, delete chat should remove chat history (and this chat history should ideally be deleted from the db and app cache); but handshake shouldn't have to be remade. 

But then you also need to block indexer rehydrating the conversation after a certain point (timestamp, DAA, whatever).

IMO deleting in your proposed fashion is very low priority for efforts, though feel free to implement it.

MEZTech-LLC:
I understand implementation is non-trivial, but seems like a pretty critical feature for a messaging application to support. 
Will try to get around to it eventually, probably requires changes to kasia-indexer.

[...]

PR #273, opened 2025-11-13
Adding image download button and configurable resizing/compression during image sending. 

MEZTech-LLC:
Title says it all. 

Reference to: https://github.com/K-Kluster/Kasia/issues/272#issue-3619149410


src/components/Modals/ImageCompressionModal.tsx
```diff
@@ -0,0 +1,263 @@
...
+    if (file) {
+      const reader = new FileReader();
+      reader.onload = (e) => setPreview(e.target?.result as string);
+      reader.readAsDataURL(file);
+    }
+    return () => setPreview(null);
+  }, [file]);
+
+  // Process image whenever settings change
+  const processImage = useCallback(async () => {
```
HocusLocusTee: We do use [React Compiler](https://react.dev/learn/react-compiler), so there would be no use for *general* 'useCallback's.

Here though, I recommend removing it and inlining the logic into the useEffect below with explicit dependencies. Worth noting a callback is not even needed, as the function is never passed anywhere and is only used in that effect.

MEZTech-LLC: Recommended changes made

HocusLocusTee:
Initial impressions (no code):

- quality scale is a great idea but you're scaling from original resolution to minimal. acceptable file sizes are in the bottom _small_ percent of any reasonable scale, making the majority of the slider redundant. it would make sense to have the auto compressed version at 100% (as its max) and allow users to scale less.
- we need some sort of de-bounce on the sliders as ever time I tried to slide it, it 'catches' and reprocesses fee estimate etc.
- there are no cursor-pointer on buttons
- can we keep the modal framing consistent and scroll within the modal? 
- probably should preview the compressed image, not the original?
- can you remove the package-lock changes?

- download image is good but needs cursor-pointers
- estimated fee etc is good

MEZTech-LLC:
@HocusLocusTee 

-So if I'm understanding correctly, select a default compression (maybe 70) and scale to maximum size allowable for default? 
Then perhaps set the upper limit on the resize slider equal to the max size allowable given the lower limit on the quality slider. 

 -Bounce probably caused by too large slider step size, should be easy fix. Will reprocess only when user "lets go" of slider.
 
 -Will add pointers
 
 -What do you mean by keeping the framing consistent and scroll within modal? Elaborate a bit please? 
 
 -Viewing compressed image, how should we handle this? If we view live resize in this settings modal, then it likely needs to adapt to fit; could have an inner modal (expansion preview) to handle this.

HocusLocusTee:
> @HocusLocusTee 
> 
> 
> 
> -So if I'm understanding correctly, select a default compression (maybe 70) and scale to maximum size allowable for default? 
> 
> Then perhaps set the upper limit on the resize slider equal to the max size allowable given the lower limit on the quality slider. 
> 
> 
> 
>  -Bounce probably caused by too large slider step size, should be easy fix. Will reprocess only when user "lets go" of slider.
> 
>  
> 
>  -Will add pointers
> 
>  
> 
>  -What do you mean by keeping the framing consistent and scroll within modal? Elaborate a bit please? 
> 
>  
> 
>  -Viewing compressed image, how should we handle this? If we view live resize in this settings modal, then it likely needs to adapt to fit; could have an inner modal (expansion preview) to handle this. 

- The current auto compression crudely finds the maximum image permissible (non optimised). So having this as the upper end of the scale - or near upper - makes sense. Though if someone wants to send an image double the size and half the quality, you wouldn't be able to.

- Currently it scrolls and you lose the X to close. Ideally that's static.

- Ideas: Keep the preview area static. Any reduction in size can just shrink inside this fixed area and increase can just expand it to a given max. Although I'm not sure who wants to send large images at immensely degraded quality. Regardless of solution, showing a preview and it not representing the compressed output seems wrong to me.

MEZTech-LLC:
Addressed:
-Missing cursors on buttons
-Consolidated compression settings (user selects image max dim and quality parameter auto selected (cleaner than having two sliders)).
-Undo package-lock changes
-Preview shown is now compressed image.

HocusLocusTee:
Hey @MEZTech-LLC 
Branch doesn't work as intended. Seems like you've deleted use of `ImageCompressionModal` from `DirectComposer`.

MEZTech-LLC:
Whoops, sorry you're right. Unintentionally removed changes to DirectComposer.
Fixed now

HocusLocusTee:
thanks for the update. Some more non-code comments:

- fee estimate does not match actual estimate (attach image and you will see _actual_ fee in fee estimate)
- what is 'quality'? Larger images have a reduced quality, smaller images have an increase quality (up to 95%). Even if accurate by some metric, I think it might be a little confusing. Maybe we should have size AND quality sliders? That makes the most sense if we are intending to give the user control over what they want.

UX concerns:
- maybe this should be behind a feature flip. By default I still think normal compression is fine.
- slider still processes on every step. ideally it doesn't 'process' until mouse / touch is released
- have you thought about tightening up the design (padding / layout) on mobile breakpoints?

MEZTech-LLC:
"fee estimate does not match actual estimate (attach image and you will see actual fee in fee estimate)"
- Noted, will fix

"what is 'quality'? Larger images have a reduced quality, smaller images have an increase quality (up to 95%). Even if accurate by some metric, I think it might be a little confusing. Maybe we should have size AND quality sliders? That makes the most sense if we are intending to give the user control over what they want."
- The "quality" is in reference to JPEG/webp quality parameters (compression degree - independent of scale). Originally I had  separate size and quality sliders, but IMO since we only support up to ~17kB images atm, we should just maximize the quality given the scale or vice-versa to fit the max allowable memory footprint (17kB is hardly enough to send images as is). 


"maybe this should be behind a feature flip. By default I still think normal compression is fine."
- I think having this behind a feature flip is reasonable, but personally I think the default compression needs to be modified/updated. Default compression should probably an image size/scale as close as possible to the original size constrained to a quality parameter above some lower bound (I believe I was using 30 for the default lower bound in my modal).  

"slider still processes on every step. ideally it doesn't 'process' until mouse / touch is released"
- Right, sorry, thought I had fixed this. I may have reverted changes unintentionally. 

"have you thought about tightening up the design (padding / layout) on mobile breakpoints?" 
- Will work on this! Haven't paid much attention to mobile styling to this point. Will give it a closer look.

HocusLocusTee:
Given such little room for compression variation, it does make me wonder if we need this feature at all.

Not sure why someone would want to send a large very low quality image. 

Posting links to images is around the corner as well.

MEZTech-LLC:
"Given such little room for compression variation, it does make me wonder if we need this feature at all."
- Yeah I'm starting to agree with this. 

"Not sure why someone would want to send a large very low quality image."
- IMO the correct path forward would simply be to alter the default compression behavior to maximize image size (up until the original image size) for a given quality lower bound. 
JPEG & webP compression "quality" is all about removing high frequency components from images. So there's actually many cases where one might want to trade quality for size (esp for conserving edge information, often necessary for legibility of text within images). 
Though I don't think there's any cases where increasing past the original image size is warranted. 
Try sending an image containing text using the default compression behavior in the branch i pushed vs the default compression behavior in the deployed branch to see what I mean here. (Or see the below images, the larger image is the default compression behavior I'm describing above (the one in my branch), the smaller is the currently deployed default compression behavior that seems to overemphasize the quality parameter over maintaining image size). 
kasia-image-2025-12-02T17-58-38
download (1)
 
 
 "Posting links to images is around the corner as well."
 - Would def be a convenient feature, but I can see cases where users would still prefer to attach the image data directly the the payload. Is there any development/consideration on sharding the image data across multiple separate transactions?  



In any case, I think including the "download" button in this branch is a good minor feature.

HocusLocusTee:
Ack on that.

Lets keep Download Button, I think its good too.
Feel free to alter this branch / PR (or start a new one) adjusting current compression settings.

MEZTech-LLC:
Changes made in this branch:

- Added download button
- Altered default compression behavior as per above.

All changes wrt ImageCompressionModal have been reverted.

src/components/Modals/ImagePresenter.tsx
```diff
@@ -16,13 +17,45 @@ export const ImagePresenter = () => {
     return null;
   }
 
+  const handleDownload = () => {
```
HocusLocusTee, line 37: 1. Lets extract this to `utils/file-download.ts`.

The current codebase is not clear (legacy reasons) but we should not be including heavy logic like this within a component. 

2. Download handling is not the nicest. On ios browser - it opens a file link which when i interact with it reloads the page and logs me out. Haven't thought of the best way to handle this.

On Tauri, we could use `if(core.isTauri())` conditional and using native file dialogs. But I find this less important. 

Or, do you think, should we disable file download feature for mobile all together? If I want to save an image I just long press and ios allows me to save to mobile. Not sure about Android.

src/components/Modals/ImagePresenter.tsx
```diff
@@ -16,13 +17,45 @@ export const ImagePresenter = () => {
...
   return (
-    <div className="flex items-center justify-center">
+    <div className="flex flex-col items-center justify-center gap-4">
       <img
         src={imagePresenterImage}
         alt="Presented image"
         className="h-auto max-w-full rounded-lg object-contain"
       />
+      <button
+        onClick={handleDownload}
```
HocusLocusTee: Something like this is less crude imo:

```
 <div className="relative inline-block">
      <img
        src={imagePresenterImage}
        alt="Presented image"
        className="h-auto max-w-full rounded-lg object-contain"
      />
      <button
        onClick={handleDownload}
        className="absolute top-1 left-1 flex cursor-pointer items-center justify-center rounded-lg bg-[var(--button-primary)]/60 p-1 text-[var(--text-primary)] transition-colors hover:bg-[var(--button-primary)]/100 active:scale-95 sm:p-2"
      >
        <Download className="size-4" />
        <span className="absolute p-4 pointer-fine:hidden" />
      </button>
    </div>
```

HocusLocusTee (changes requested):
There's a lot of redundancy and repetition in `upload-file-service`.
Maybe (to name a few) we could:
- extract some of the repeated patterns
- break apart the compressImageToFit function
- reorg some of the constants

src/service/upload-file-service.ts
```diff
@@ -89,84 +91,261 @@ export async function prepareFileForUpload(
   }
 }
 
-// Try multiple times to compress. reduce the quality for the first half of attempts, then reduce both for the remaining
+// Constants for compression logic matching ImageCompressionModal
```
HocusLocusTee: referencing modal that does not exist.

src/components/Modals/ImagePresenter.tsx
MEZTech-LLC: 1. Done

2.  I've attempted a fix in file-download.ts whereby we include the file link in the DOM to prevent a page reload on IOS.
I don't have an easy debug env to check if this works or not, so let me know.

src/components/Modals/ImagePresenter.tsx
MEZTech-LLC: Awesome, I like it

src/service/upload-file-service.ts
MEZTech-LLC: Thanks

src/service/upload-file-service.ts
HocusLocusTee, line 1: We could clean up with:

```
export interface CompressImageOptions {
  minWidth?: number;
  minHeight?: number;
  maxQuality?: number;
  minQuality?: number;
  qualityStep?: number;
}
```

and then

```
const DEFAULT_COMPRESSION_CONFIG: CompressImageOptions = {
  minWidth: 150,
  minHeight: 150,
  maxQuality: 0.95,
  minQuality: 0.3,
  qualityStep: 0.05,
};
```

src/utils/file-download.ts
```diff
@@ -0,0 +1,57 @@
+/**
+ * Safely trigger a file download that works across browsers including iOS Safari.
+ *
+ * The key difference from a simple link.click() is that we:
+ * 1. Append the link to the DOM before triggering (required for iOS)
+ * 2. Remove it after a delay to ensure the download starts
+ * 3. This prevents iOS Safari from navigating to the link and reloading the page
+ */
+export function downloadFile(dataUrl: string, fileName: string): void {
```
HocusLocusTee, line 22: I think safari just sucks. This method is no better for tab navigation (utility abstraction is good though).
Its just better to dataUrlToBlob and then just url the blob and anchor tag it. < close to what you were doing before.

^ I've tried this and it doesn't re-load the page. Lmk if you want me to commit my changes.

HocusLocusTee:
Apologies for the back and forth. We spent a lot of time trying to de-clutter the codebase from the original PoC, and much of the code is still far from ideal. When we get the chance to implement anything, we ideally refactor + try to set best practices at the same time.

src/service/upload-file-service.ts
MEZTech-LLC: Done

src/utils/file-download.ts
MEZTech-LLC: Added a dataUrlToBlob method; let me know if that works on your end - if it doesn't feel free to commit your changes

MEZTech-LLC:
No worries, totally understand!

HocusLocusTee:
Hey @MEZTech-LLC , 

This is great! Mobile download works well (on ios). I'm very impressed with the quality you're managing to pack in. It looks astronomically better.
I've pushed a commit to this branch (permission denied on your branch) https://github.com/K-Kluster/Kasia/pull/282

Just some minor styling updates, so feel free to check that out and cherry pick to yours (maybe squash some of the middle commits too? no big deal though) or I can just merge it after this is merged.

Anyway, PR looks good to me. Thanks so much for the work and attention to detail.

HocusLocusTee (approved):
See other comment

MEZTech-LLC:
Merged your changes into this branch. 
Thanks for all your help!

PR #275, opened 2025-11-28
Chore/localstorage clean

HocusLocusTee:
## what?
deletes old keys with patterns that leaked privacy. replaces 'last contact' and 'last bcast channel' with safe versions:
```
"kasia_last_opened_recipient_{address}"
"kasia_last_opened_contact_id_{address}"
"kasia_last_opened_channel_id_{address}"
"encrypted_conversations_kaspa"
"kaspa_messages_by_wallet"
```
_we don't need to worry about migrating the above. its fine._

removes all db migration logic and deletes local storage keys (all wallets) and all code
- this was creating a new migration flag for each wallet (new or old)

note: since we aren't planning to release any time soon, migrations should all be 'done' by the time we release this

- delete active local storage stuff when wallet is deleted
- bump version to 0.6.3 because i reference it in a comment. we wont need to bump again prior to our next release unless there is something major

## test
- once you boot up all the following keys should be deleted from local storage (note it took 2 logins for me to delete all the migration key entries.
- create a wallet, see the new keys, delete wallet make sure they delete.


IzioDev:
We can rely on `metadata` version from now on to run safe migrations version to version (and increment accordingly after the migration). This would give us more confidence in the future to avoid removing / adding new migrations without keeping track of the current version

But this isn't the scope of this PR, so if you want to merge it regardless, I'm fine :)

HocusLocusTee:
> We can rely on `metadata` version from now on to run safe migrations version to version (and increment accordingly after the migration). This would give us more confidence in the future to avoid removing / adding new migrations without keeping track of the current version
> 
> But this isn't the scope of this PR, so if you want to merge it regardless, I'm fine :)

Ack. We can use this going forward. 
Since we are both ok with current hard removal - lets keep it as is for this PR.

PR #276, opened 2025-11-29
Chore/package bumps + rendering fix

HocusLocusTee:
## what?
- audit fix, pining some package versions
- bump react compiler to v1.0
- clean some minor redundant comments

- also fixed a bug where we were getting infinite rendering loops on broadcasts. this sometimes caused a bug where you couldn't 'switch' back to directs.


IzioDev:
lgtm, very good to fix versions

PR #277, opened 2025-12-01
Feat/markdown rendering

HocusLocusTee:
## What?
adds markdown rendering to messages

by default, all clients will render Markdown messages if they include the `MARKDOWN_PREFIX = "§"`. However, to send markdown messages, you must enable it via a feature flip. This is because of the higher cost etc.

with rendering enabled, broadcasts do not render links or images by default. This must be enabled separately as a sub-feature. In direct messages, links and images are always rendered.

currently we (I) are only planning to support the base CommonMark standard. As I *personally* don't think a chat app needs rich github level markdown features. There is also no UI interface to interactively place markdown or to see the markdown as you write it, maybe this can happen in future versions (item 2 here being of more importance than item 1) .

image

### DIFF
--- a/src/components/MessagesPane/Utilities/Content/MarkdownRenderer.tsx
+++ b/src/components/MessagesPane/Utilities/Content/MarkdownRenderer.tsx
@@ -38,7 +38,11 @@ export const MarkdownRenderer: React.FC<MarkdownRendererProps> = ({
     <ReactMarkdown
       remarkPlugins={[remarkBreaks]}
       components={{
-        a: ({ href, children, ...props }) =>
+        a: ({
+          href,
+          children,
+          ...props
+        }: React.AnchorHTMLAttributes<HTMLAnchorElement>) =>
           enableMdLinks ? (
             isSafeHref(href) ? (
               <a
@@ -60,16 +64,21 @@ export const MarkdownRenderer: React.FC<MarkdownRendererProps> = ({
               </a>
             ) : (
               <span {...props}>
-                UNSAFE LINK | Link Text: {children} | Link: {href}
+                Kasia Safety Guard - Not Standard Link | Link Text: {children} |
+                Link: {href}
               </span>
             )
           ) : (
             <span {...props}>
-              KASIA PERMISSION - Links Not Enabled | Link Text: {children} |
+              Kasia Safety Guard - Links Not Enabled | Link Text: {children} |
               Link: {href}
             </span>
           ),
-        img: ({ src, alt, ...props }) =>
+        img: ({
+          src,
+          alt,
+          ...props
+        }: React.ImgHTMLAttributes<HTMLImageElement>) =>
           enableMdImages ? (
             <img src={src} alt={alt} {...props} />
           ) : (
--- a/src/components/MessagesPane/Utilities/Content/MarkdownRenderer.tsx
+++ b/src/components/MessagesPane/Utilities/Content/MarkdownRenderer.tsx
@@ -3,10 +3,14 @@ import ReactMarkdown from "react-markdown";
 import rehypeSanitize, { defaultSchema } from "rehype-sanitize";
 import remarkBreaks from "remark-breaks";
 
-// extend default schema to ensure br tags are allowed
+// extend default schema to ensure br tags are allowed and links can open in new tabs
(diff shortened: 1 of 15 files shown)


src/components/MessagesPane/Utilities/Content/MarkdownRenderer.tsx
```diff
@@ -0,0 +1,58 @@
...
+          enableMdImages ? (
+            <img src={src} alt={alt} {...props} />
+          ) : (
+            <span className="text-gray-500 italic" {...props}>
+              [KASIA PERMISSION - Image Links Not Enabled: {src}]
+            </span>
+          ),
+      }}
+    >
+      {content}
```
IzioDev, line 91: consider sanitize before rendering: https://developer.mozilla.org/en-US/docs/Web/API/HTML_Sanitizer_API

IzioDev:
good idea to use a prefix, this keeps the message coupled with the used format

src/components/MessagesPane/Utilities/Content/MarkdownRenderer.tsx
HocusLocusTee: > consider sanitize before rendering: https://developer.mozilla.org/en-US/docs/Web/API/HTML_Sanitizer_API
> 

Previous commits use sanitization but I removed as I don't really think it's necessary on this case. Lmk if you'd like me to add it again.

IzioDev: Why do you think it is not necessary?

I'm considering the following case:
* I create a malicious tag that render an image, `` (or any other uncontrolled DOM attack vector)

HocusLocusTee: On mobile atm so excuse formats and general links.

https://github.com/remarkjs/react-markdown

React markdown escapes html. We are also not using 'rehype-raw'.

I can add 'skipHtml' - but that would remove it all together and potentially destroy code blocks? (A guess, not tested).

Or I can add sanitization.

IzioDev: I'm not ultra familiar with `react-markdown` internals, but I can see on their README they state to use a plugin to sanitize.

image

[source](https://github.com/remarkjs/react-markdown)

I assume if they recommend this particular plugin, it is because it doesn't break the structure? (to be validated)

HocusLocusTee: All good, we can err on safety. I recently removed it so it shouldn't be an issue to add it back in.
Will ping you when I do.

IzioDev: Awesome thanks, I'll try it out then :)

Well needed feature! 🫡

HocusLocusTee: @IzioDev I’m inclined to stick with not adding sanitization for now.

As per the react-markdown security notes (image you shared), the default renderer escapes/ignores raw HTML so raw HTML in markdown doesn’t become DOM nodes unless we add something like `rehype-raw`. 

Our only plugin is remark-breaks, which [states ](https://github.com/remarkjs/remark-breaks?tab=readme-ov-file#security)it doesn’t introduce XSS openings. 

I’m not against sanitization on principle, but my concern is UX and robustness(?) e.g. In my testing, enabling sanitization causes malformed raw HTML (eg a harmless `<script>` with no closing tag) to swallow the rest of a message.

But a strong agree that if we plan to enable raw HTML or add more rehype plugins, we should revisit `rehype-sanitize` with an explicit schema. I've also added a comment in the component for this.

Happy to change course if there’s a concrete exploit (or convincing tradeoff) in the current setup.

HocusLocusTee: Added a small allow list to sanitize md links. But beyond that - without rehype raw - I still think there is no concern.

IzioDev:
Very small comment, maybe replace [KASIA PERMISSION - xyz] with `Kasia Safety Guard` or something similar?
What do you think?

Mergeable, looks great, working good

PR #278, opened 2025-12-06
Feat/direct send status

HocusLocusTee:
## what?
adds 'pending, active, failed' message status to direct messages and payments.
follows the current broadcast approach.

handshakes have _not_ been done as they are a little more complicated.

followup tasks from this:
- add custom fees to 'payments' (custom fees should exist for wallet withdrawals and handshakes as well, but payments should have some urgency due to its UX)
- extend status into handshakes
- add rbf for pending tx
- refactor message ingestion once vccv2 is out in the real world

### DIFF
--- a/src/components/MessagesPane/Directs/DirectDisplay.tsx
+++ b/src/components/MessagesPane/Directs/DirectDisplay.tsx
@@ -127,7 +127,10 @@ export const DirectDisplay: FC<DirectDisplayProps> = ({
             generateBubbleClasses({
               isOutgoing,
               groupPosition,
-              status: event.__type === "message" ? event.status : "confirmed",
+              status:
+                event.__type === "message" || event.__type === "payment"
+                  ? event.status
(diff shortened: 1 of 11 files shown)


IzioDev:
Working good, identified one flaw which is minor and can be ignored:
- If I quit the application while having a pending message and I re-open, it never updates

Knowing this, is it necessary to persist in DB the status?

In any ways you want to proceed with this, mergeable

HocusLocusTee:
> Working good, identified one flaw which is minor and can be ignored:
> 
>     * If I quit the application while having a pending message and I re-open, it never updates
> 
> 
> Knowing this, is it necessary to persist in DB the status?
> 
> In any ways you want to proceed with this, mergeable

Thanks for the heads up.
I've created an issue card  with proposed solution. https://github.com/K-Kluster/Kasia/issues/292

Note: I think its important to keep status. As we might need it for RFB etc later on.

PR #281, opened 2025-12-16
refactor: wallet storage service

HocusLocusTee:
## what?
refactors how we generate and handle public and private keys.
there was some incorrect terminology in the wallet-storage-service. (e.g. calling `encryptedXPrv` when it was in fact the seed). This lead me to have a look and refactor some things for clarity.

- removes the public key generator, stops assigning the wallets seed against the wallet object
- replaces the above with the derived private key at index 0(we still obviously keep the mnemonic in storage)
- I've also reduced the props that were being passed to our privatekey return

note: this works well under the assumption we only use one address / key set. at the moment this is the current state of things.
I don't believe this change will make any future work more difficult.

## testing
everything should work as is. nothing changes. no migrations needed.

### DIFF
--- a/src/service/wallet-storage-service.ts
+++ b/src/service/wallet-storage-service.ts
@@ -2,8 +2,8 @@ import {
   decryptXChaCha20Poly1305,
   encryptXChaCha20Poly1305,
   Mnemonic,
+  PrivateKey,
   PrivateKeyGenerator,
-  PublicKeyGenerator,
   XPrv,
 } from "kaspa-wasm";
 import { v4 as uuidv4 } from "uuid";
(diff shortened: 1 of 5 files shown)


src/service/wallet-storage-service.ts
```diff
@@ -78,27 +78,30 @@ export class WalletStorageService {
         );
       }
 
-      // Generate the seed with passphrase and extended private key
+      // Generate the seed with passphrase and create master extended private key
```
IzioDev, line 81: No change, just a general question: is `xPrv` (extended private key) and `master` private key the exact same in term of representation?

src/store/messaging.store.ts
```diff
@@ -483,11 +479,7 @@ export const useMessagingStore = create<MessagingState>((set, g) => {
       }
 
       // cannot use `walletStore.address` because it is not always already populated
-      const address = WalletStorageService.getPrivateKeyGenerator(
-        unlockedWallet,
-        unlockedWallet.password
-      )
-        .receiveKey(0)
```
IzioDev, line 490: Side: we could even store the receive address upon unlocking since we use only one (and they we can remove these dynamically constructed address). But that's out of scope

IzioDev:
LGTM

src/service/wallet-storage-service.ts
HocusLocusTee: Not sure if I follow exactly, so ill try to be explicit...
We derive the masterXPrv key from the seed (this is extended and not just a raw private key.
We then use masterXPrv to derive the usable pk which is Account(0)Pk(0)
There's no handling of the 'raw' master pk.
Is that what you were looking for @IzioDev

src/store/messaging.store.ts
HocusLocusTee: Good idea! Leave this as a future task.

src/service/wallet-storage-service.ts
IzioDev: 100% 👍

PR #283, opened 2025-12-17
Deterministic Alias Assignment

MEZTech-LLC:
Changes for deterministic alias assignment. 
See notes in docs/deterministic_aliases

Also note that proto file in kasia-indexer should be changed to reflect a unified conversation alias.


HocusLocusTee:
FYI: Instead of `conversation/s` I've been trying to move the terminology for 1:1 messages to `direct/s`.

I envision that conversation covers all non-broadcast messages (directs and groups).

docs/deterministic_aliases/CHANGELOG.md
HocusLocusTee, line 1: Note sure what the solution is, but not a fan of adding all these docs (>1) to the repo. If necessary, this could be a  catalyst for a better  solution.

PR #284, opened 2025-12-23
Deterministic Aliases

MEZTech-LLC:
New implementation (deterministic, but non-unified aliases). 
Distilled related documentation.


HocusLocusTee:
Functional test - seems to work as intended.

You'll need to revise the 'Offline Handshake' flow though as it's now broken.
How to access: Long hold on the new conversation + symbol.

Whether we remove this and absorb it as part of the discrete mode - makes sense) - but this hidden mode is far 'cleaner' than the current new chat window (speaking UI here). So maybe some consolidation needs to be done.

Also, adding a discrete chat should probably still show a message or something that handshake is completed? Maybe something like "Discrete Chat Initiated".

Also, I'm still not sold on these docs being part of the repo.

HocusLocusTee:
^ Edited above.

HocusLocusTee:
Also FYI, maybe a little more detail in the PR desc next time.

MEZTech-LLC:
**"You'll need to revise the 'Offline Handshake' flow though as it's now broken.
How to access: Long hold on the new conversation + symbol.

Whether we remove this and absorb it as part of the discrete mode - makes sense) - but this hidden mode is far 'cleaner' than the current new chat window (speaking UI here). So maybe some consolidation needs to be done."**

- Wasn't aware of that flow. Somewhat disagree that the hidden mode is cleaner. It's quite difficult to discover despite being a pretty important feature (imo). What do you recommend in terms of merging the two? My opinion is we should just remove the hidden mode and keep the optional discrete convo checkbox, but open to leaving the "hidden option"/existing offline handshake flow if we can figure out a clean way to make it more discoverable. 



**"Also, adding a discrete chat should probably still show a message or something that handshake is completed? Maybe something like "Discrete Chat Initiated"."**
- Roger that

**"Also, I'm still not sold on these docs being part of the repo."**
- Sure, can remove. I mostly left them in there just to help with debugging as we add this feature.

HocusLocusTee:
> 
> 
> **"You'll need to revise the 'Offline Handshake' flow though as it's now broken.
> 
> How to access: Long hold on the new conversation + symbol.
> 
> 
> 
> Whether we remove this and absorb it as part of the discrete mode - makes sense) - but this hidden mode is far 'cleaner' than the current new chat window (speaking UI here). So maybe some consolidation needs to be done."**
> 
> 
> 
> - Wasn't aware of that flow. Somewhat disagree that the hidden mode is cleaner. It's quite difficult to discover despite being a pretty important feature (imo). What do you recommend in terms of merging the two? My opinion is we should just remove the hidden mode and keep the optional discrete convo checkbox, but open to leaving the "hidden option"/existing offline handshake flow if we can figure out a clean way to make it more discoverable. 
> 
> 
> 
> 
> 
> 
> 
> **"Also, adding a discrete chat should probably still show a message or something that handshake is completed? Maybe something like "Discrete Chat Initiated"."**
> 
> - Roger that
> 
> 
> 
> **"Also, I'm still not sold on these docs being part of the repo."**
> 
> - Sure, can remove. I mostly left them in there just to help with debugging as we add this feature. 

I should've specified... I don't think the 'hidden-ness' is cleaner - rather the model itself.

MEZTech-LLC:
Alright, 
Perhaps having tabs for each convo type at the top of the modal is better then? (Handshake and Offline-Handshake) where the offline handshake tab shows the existing modal?

HocusLocusTee:
> Alright, 
> 
> Perhaps having tabs for each convo type at the top of the modal is better then? (Handshake and Offline-Handshake) where the offline handshake tab shows the existing modal?

Decent idea. Maybe let's leave it as is, do the clean up and the review of core idea and after I'm happy to propose some UI work.

MEZTech-LLC:
@HocusLocusTee 
Made changes discussed above + added a fix for an edge case handshake logic bug. 
If user A and B had an ongoing discrete chat and user A deleted their conversation, if user A attempts to "rehandshake" (non-discrete) with B, user A would not be able to receive messages from B due to no handshake response being received. 
Changed behavior to automatically send a handshake response for any chat that is currently "active" in response to a handshake attempt.

HocusLocusTee:
Part of the discrete conversation flow should give the option to self stash the alias.

Also, self stash or not - from a UX perspective we should add a self (off-chain) message saying a discrete handshake was completed.

This will also add a date time so we don't get a blank card with no reference like the image below:

image

MEZTech-LLC:
Ok added a toast for the discrete message init, but will change it to a message in the thread.

What do you mean by "self stash the alias"?

HocusLocusTee:
>What do you mean by "self stash the alias"?

In the current system we send a 'self stash' tx to ourselves - which is indexed by the indexer - so we can keep the handshake between difference devices.

Now, in the proposed alias changes: if I am `A` and set up a discrete handshake with `B`, then chance devices -  we will need to self stash this to sync it.

MEZTech-LLC:
@HocusLocusTee 
Added proposed changes. 

Also something I noticed that would be a useful future improvement; seems like contact names/nicknames are not stored anywhere except the local db. Should probably track these on the indexer using some sort of similar self-stash method.

HocusLocusTee:
1. If A has no funds, they cannot start Discrete chat. (They should be able to). We should find a way to  segregate the self stash tx and the creation of chat. See old Offline Handshake flow.
image

2. Still trying to think of a better way to handle new chats. This branch currently adds a 'Discrete Chat Initiated' message but the in chat link for that is now a dead-link. `https://explorer-tn10.kaspa.org/txs/discrete_start_b2bcf46f-2df1-46df-b9d7-c04e317411f6`. We could do a hacky patch but there might be a better option?

RE: your statement: Sure, seems low priority. But IMO it's outside of this scope. Maybe open an issue ticket for it.

MEZTech-LLC:
"If A has no funds, they cannot start Discrete chat. (They should be able to). We should find a way to segregate the self stash tx and the creation of chat. See old Offline Handshake flow."

Not sure I understand this. 
Doesn't self-stashing require the user to have at least some funds, otherwise they wouldn't be able to conduct a tx (cover fees). 
Do you just mean that the 0.2 KAS balance shouldn't be required? Or do you mean that the .2 KAS balance shouldn't be required AND if the balance is 0 (not enough to cover the network fee) the self-stash should just be skipped?

HocusLocusTee:
Self stash does require KAS. But setting up a discrete handshake does not.

Right now, no balance blocks me from even creating the handshake.

If balance is 0, we could automatically skip the self stash... but we should _eventually_ add a way to stash it again at a later date.

MEZTech-LLC:
Got it - self stash is now skipped when funds are not available.

src/store/messaging.store.ts
```diff
@@ -1307,18 +1313,35 @@ export const useMessagingStore = create<MessagingState>((set, g) => {
...
-      // Create self-stash for discrete conversation to sync across devices
-      const selfStashTxId = await g().createSelfStash({
-        type: "initiation",
-        partnerAddress: recipientAddress,
-        ourAlias: conversation.myAlias,
-        theirAlias: conversation.theirAlias ?? undefined,
-      });
+      // Optionally create self-stash for cross-device sync (requires minimal funds for network fees)
+      // Check if user has sufficient balance before attempting
+      const minFeeAmount = BigInt(100000); // ~0.001 KAS for network fees
```
HocusLocusTee: we use  `const minAmount = kaspaToSompi("0.19");` in other places. maybe we should generalize this.

src/components/Layout/ModalHost.tsx
```diff
@@ -13,7 +13,6 @@ import { LoaderCircle } from "lucide-react";
 import { ImagePresenter } from "../Modals/ImagePresenter";
 import { BroadcastParticipantInfo } from "../Modals/BroadcastParticipantInfo";
 import { QrScannerModal } from "../Modals/QrScannerModal";
-import { OffChainHandshakeModal } from "../Modals/OffChainHandshakeModal";
```
HocusLocusTee, line 17: We need to clean up `OffChainHandshakeModal`.

src/components/SideBarPane/HoldablePlusButton.tsx
```diff
@@ -16,40 +14,29 @@ export const HoldablePlusButton: FC<HoldablePlusButtonProps> = ({
   broadcastEnabled,
   isBroadcastMode,
   onNewChat,
-  onOffChainHandshake,
   onNewBroadcast,
   collapsed = false,
 }) => {
-  const handleShortPress = () => {
-    onNewChat();
+  const handleClick = () => {
```
HocusLocusTee: We don't need this component at all anymore. It can be styled with a regular + button and none of the 'holdable' complexity.

DEBUGGING_ALIAS_ROUTING.md
```diff
@@ -0,0 +1,306 @@
+# Debugging Alias Routing - Logging Added
```
HocusLocusTee: I still dont think we need this doc. There are plenty of decent comments explaining how this works in the new util file.

src/store/messaging.store.ts
```diff
@@ -1154,6 +1261,97 @@ export const useMessagingStore = create<MessagingState>((set, g) => {
...
+          );
+        } catch (error) {
+          console.warn(
+            "[createDiscreteConversation] Failed to create self-stash (conversation still works locally):",
+            error instanceof Error ? error.message : error
+          );
+        }
+      } else {
+        console.log(
+          "[createDiscreteConversation] Skipping self-stash - insufficient balance for network fees (conversation works locally)"
```
HocusLocusTee, line 1276: Great. But how does one self stash at a later date?
We should add a warn that we did not self stash due to no balance.
We also need to add the ability to self stash at a later date (and track this somehow). I don't think it should be part of this PR as it'll muddle the context - but we should be aware and open an issue /  put it into future work.

src/store/messaging.store.ts
```diff
@@ -1414,12 +1616,12 @@ export const useMessagingStore = create<MessagingState>((set, g) => {
         throw new Error("Account service not available");
       }
 
-      // check if user has sufficient funds (0.2 KAS minimum)
-      const minAmount = BigInt(20000000);
+      // check if user has sufficient funds for network fees (not 0.2 KAS like handshakes)
+      const minFeeAmount = BigInt(100000); // ~0.001 KAS for network fees
```
HocusLocusTee: Same comment as above... but what was the reasoning to change this?

HocusLocusTee (changes requested):

src/components/SideBarPane/HoldablePlusButton.tsx
MEZTech-LLC: I had left it for future usecases, but will remove

src/store/messaging.store.ts
MEZTech-LLC: Want me to open this as a new  issue? Or can you?

src/store/messaging.store.ts
MEZTech-LLC: The reason I changed from 0.2 KAS to .001 KAS is because it didn't appear logical to require a minimum KAS amount  that was substantially greater than common transaction fees. 

I may have been missing the intent though. 
Is the intent of requiring 0.2 KAS to prevent spam attacks on the indexer?

HocusLocusTee: Is nearly infeasible to get a wallet with 0.001.
It could happen if I start with an amount that is small and send enough messages whittling away my balance. Then I guess only 1:1 messages would be able to be spent.

But I think we keep it as 0.2 kas and if you feel strongly - lets open an issue to do some RnD on kasia features and KIP-9 interactions. Additionally, we don't have fee setting for handshake or wallet withdrawals - so I think they should all come first.

src/store/messaging.store.ts
MEZTech-LLC: Leaving this as is for now - linter flags usage in the subsequent conditional if I change it to use kaspaToSompi since undefined can be returned.

HocusLocusTee:
@MEZTech-LLC please see the commits in this branch:
https://github.com/K-Kluster/Kasia/pull/286

Feel free to cherry-pick them if you approve.

src/store/messaging.store.ts
HocusLocusTee: https://github.com/K-Kluster/Kasia/issues/287

MEZTech-LLC:
Looks good to me

src/store/messaging.store.ts
```diff
@@ -1102,7 +1124,99 @@ export const useMessagingStore = create<MessagingState>((set, g) => {
...
+        result &&
+        typeof result === "object" &&
+        "shouldSendResponse" in result &&
+        "conversationId" in result
+      ) {
+        const typedResult = result as {
+          shouldSendResponse: boolean;
+          conversationId: string;
+        };
+        if (typedResult.shouldSendResponse) {
```
HocusLocusTee: I think we should extract some of this to the conversation manager stop. Maybe something like `sendAutomaticHandshakeResponse`

src/store/messaging.store.ts
```diff
@@ -1116,27 +1230,22 @@ export const useMessagingStore = create<MessagingState>((set, g) => {
       const manager = g().conversationManager;
       return manager ? manager.getPendingConversationsWithContact() : [];
     },
-    // add an offline handshake
+
+    /**
+     * @deprecated Manual alias exchange is no longer needed. Aliases are derived deterministically.
```
HocusLocusTee: Probably dont need legacy comments like this and the one on line 1247

src/service/block-processor-service.ts
```diff
@@ -293,6 +304,15 @@ export class BlockProcessorService extends EventEmitter<{
       this.monitoredAddresses.clear();
       const conversations = conversationManager.getMonitoredConversations();
 
+      console.log("[block-processor] Updating monitored aliases:", {
```
HocusLocusTee: I think we should remove the logs you've added. They are way too noisy.

HocusLocusTee: @MEZTech-LLC I've done it for you. feel free to just pick the last 2 commits so I don't claim write ownership over your other commits
https://github.com/K-Kluster/Kasia/pull/289

MEZTech-LLC: Thanks @HocusLocusTee - Apologies I couldn't get to this sooner; other pressing tasks came up

cipher/src/lib.rs
```diff
@@ -332,6 +334,111 @@ pub fn decrypt_with_secret_key(
...
+/// # Returns
+/// A 12-character hex string (6 bytes) representing my alias
+#[wasm_bindgen]
+pub fn derive_my_alias(
+    my_private_key: WalletPrivateKey,
+    their_address: &str,
+) -> Result<String, JsError> {
+    // Get my public key from my private key
+    let my_public_key = my_private_key.to_public_key()
+        .map_err(|_| JsError::new("Failed to derive public key"))?;
```
IzioDev: Minor: We can avoid computing pk from sk if the client provide a public key instead of the private key.
Public key can safely be computed once by the client upon startup (where we initiate the wallet state), and provide it here.
same for xonly_pk (below)

cipher/src/lib.rs
```diff
@@ -332,6 +334,111 @@ pub fn decrypt_with_secret_key(
...
+    my_private_key: WalletPrivateKey,
+    their_address: &str,
+) -> Result<String, JsError> {
+    // Parse their address to extract their public key
+    let address = Address::try_from(their_address)
+        .map_err(|e| JsError::new(&format!("Invalid address: {}", e)))?;
+    
+    // Extract X-only public key from address (32 bytes, no parity)
+    let their_xonly_pk = XOnlyPublicKey::from_slice(address.payload.as_slice())
+        .map_err(|e| JsError::new(&format!("Invalid public key in address: {}", e)))?;
```
IzioDev: This block is done twice, once here and once in `derive_alias_with_context`.

I propose to let responsibility to calling site to compute address + xonly_pk and now `derive_alias_with_context` takes a `xonly_pk` as argument

cipher/src/lib.rs
```diff
@@ -332,6 +334,111 @@ pub fn decrypt_with_secret_key(
...
+    // Convert my private key to k256 SecretKey
+    let my_sk = SecretKey::from_slice(&my_private_key.secret_bytes())
+        .map_err(|_| JsError::new("Invalid private key"))?;
+    
+    // Perform ECDH to get shared secret
+    let shared_secret = diffie_hellman(my_sk.to_nonzero_scalar(), their_pk.as_affine());
+    
+    // Construct info string: "chat" || shared_secret || context_pubkey
+    let mut info = Vec::new();
+    info.extend_from_slice(b"chat");
```
IzioDev: If we re-use another domain separation, maybe we should name `chat` more explicitly.

What it is, is: a 1:1 conversation alias for now. I propose to simplify as `alias` as it may be a general concept within the application for other purpose as-well

IzioDev (changes requested):
Rust part sounds good, I'll dedicate more time on the application code later (and actually try it locally too :) ).

cipher/src/lib.rs
IzioDev: or `dm_alias`, or `pchat` (private chat) or w/e

nit: could be made under an `enum`: enum HashDomain { ... }

src/service/conversation-manager-service.ts
```diff
@@ -736,12 +965,13 @@ export class ConversationManagerService {
...
-          monitored.push({
-            alias: conversationAndContact.conversation.theirAlias,
-            address: conversationAndContact.contact.kaspaAddress,
-          });
-        }
+        // CRITICAL: Monitor myAlias (not theirAlias)
+        // In deterministic system: I monitor myAlias, they send to theirAlias
+        // Due to ECDH symmetry: their theirAlias === my myAlias
+        monitored.push({
+          alias: conversationAndContact.conversation.myAlias,
```
HocusLocusTee, line 938: This is correct, but will be an issue with old v1 alias. Do we just optimistically hope they upgrade? If so, why do we then even give them the option to upgrade if the code isn't fully compatible with v1.

src/components/MessagesPane/Directs/ConversationUpgradeNotice.tsx
```diff
@@ -0,0 +1,161 @@
...
+        {mode === "respond" && (
+          <>
+            <p>
+              This contact requested a conversation upgrade. Send a handshake
+              response to finish the upgrade.
+            </p>
+            <Button
+              variant="primary"
+              onClick={sendUpgradeResponse}
+              disabled={isSubmitting}
```
HocusLocusTee: Maybe we should disable this with a funds check and a message if they try to send with no funds?

HocusLocusTee: General Q for this great feature: Wonder if we need a way to 'force' an upgrade. Lets say they are not getting upgrade messages, we *can* compute this without interaction, so should we allow it?

src/service/conversation-manager-service.ts
IzioDev: True, maybe just lock the conv and force manual upgrade (no hidden / automatic mechanism), wdyt?

HocusLocusTee: Is it with just running a db migration on boot?

IzioDev: Currently, on boot it adds a version field at conversation level (the entity holding alias data) and we set it to v1 by default.
Later, when you enter a conversation, if it's v1, you're asked to migrate (it replays handshake, and set to v2).
All new conversation are created as v2

HocusLocusTee: Ack on that. But why not just compute new alias on db bump / opening new version?

IzioDev: Absolutely right, 0 need into re-initiating an handshake. Besides it would reveal stealth conversation.

IzioDev: Should be good with latest commit, let me know what you think, silently migrate basically

src/components/MessagesPane/Directs/ConversationUpgradeNotice.tsx
IzioDev: Removed because I don't think it has a use-case at the moment, maybe for dcom upgrade (not even sure that'll be necessary)

PR #288, opened 2026-01-24
feat: add testnet-12 support

HocusLocusTee:
## what?
adds tn-12
can manually use node:
`wss://tn12-wrpc.kasia.fyi`

no indexer available

faucet here:
https://faucet-tn12.kaspanet.io/


IzioDev:
Perfect!
There will be a network reset in the upcoming days I think given the current velocity, but I'll keep domain consistency while upgrading.

Mergeable

PR #291, opened 2026-01-30
Chore/wallet terminology

HocusLocusTee:
## what?
updates the use of the term 'wallet' to 'account' in user facing components.
mainly to lower the conceptual barrier and not to lead people into thinking kasia should be used as a wallet.

note: not worth doing this to our internal components. I think its a fine abstraction.


KaspaSilver:
Totally agree with this update here!

IzioDev:
Neat, I didn't run locally, but it looks good to me.
It's very fair to make this abstraction at UI level and keep our current structure

HocusLocusTee:
Ty for the feedback. Ill merge it as it wont break anything and we can just test it through testing other PR that are rebased on this.

PR #294, opened 2026-06-05
Chore/dep upgrades

HocusLocusTee:
## what?

- bumps js packages
- rewords docs to use v2.0 wasm
- modifies explorer links
- updates feel colors

### DIFF
--- a/src/components/PriorityFeeSelector.tsx
+++ b/src/components/PriorityFeeSelector.tsx
@@ -172,6 +172,22 @@ export const PriorityFeeSelector: FC<PriorityFeeSelectorProps> = ({
     return buckets;
   }, [feeEstimate]);
 
+  const baseFeeRate = dynamicFeeBuckets[0]?.feerate ?? 1;
+
+  const formatFeeRateLabel = (feerate?: number) => {
+    if (feerate == null || baseFeeRate <= 0) {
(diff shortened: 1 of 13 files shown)


src/components/PriorityFeeSelector.tsx
```diff
@@ -172,6 +172,22 @@ export const PriorityFeeSelector: FC<PriorityFeeSelectorProps> = ({
...
+  const formatFeeRateLabel = (feerate?: number) => {
+    if (feerate == null || baseFeeRate <= 0) {
+      return "Base fee";
+    }
+
+    const normalizedRate = feerate / baseFeeRate;
+
+    if (Math.abs(normalizedRate - 1) < 0.001) {
+      return "Base fee";
+    }
```
IzioDev: Are we making the assumption than node is acting badly here?

IzioDev (approved):
Looks good, just one question :)

src/components/PriorityFeeSelector.tsx
HocusLocusTee: This is fixed. I noticed I was also calculating fee rate wrong. I was calculating and displaying the x representation as how many multipliers over the low bucket. Not how many multipliers above the true base fee.

